Issue #32, opened 2021-06-24
[BUG?] MACRO CHOICE: Workbench Plug-in Hotkeys  (0.2.10)


FelipeRearden:
Hello @chhoumann !!!

First of all, I just wanna thank you for everything that you have done for this community. All of yours plugins are a game-changer. 

I'm just getting started with Quickadd and my first road is about the MACRO CHOICE. 
- I'm using MACRO to automate my hotkeys workflow. Every workflow that I use more than one hotkey will be "transferred" to Quickadd.

I found one specific situation using 2 hotkeys from the [Workbench plug-in](https://github.com/ryanjamurphy/workbench-obsidian/issues) that I wanna share with you.

The HotKeys are: 
1 `Workbench: Copy the Current Line/Block to your Workbench`
2 `Workbench: Link the Current Line/Block into your Workbench`

-> The first hotkey copy the current line to the workbench note 
-> The second create a blockrefence and copy the path into the workbench note

I created a MACRO CHOICE and a MACRO like this:
![Screen Shot 2021-06-24 at 15 05 19](https://user-images.githubusercontent.com/85180696/123316860-b8fae280-d503-11eb-8755-feb2540886ad.png)

![Screen Shot 2021-06-24 at 15 08 10](https://user-images.githubusercontent.com/85180696/123316934-d334c080-d503-11eb-9770-3383d7275a24.png)

When I activate the Quickadd from the prompt, ONLY the SECOND hotkey of the MACRO is lanched
- As you can see, the 2 hotkeys are running, but only the second is working
![Screen Shot 2021-06-24 at 15 13 35](https://user-images.githubusercontent.com/85180696/123317080-fb242400-d503-11eb-9c2d-2c00fe49123d.png)

- Doing this same operation without Quickadd works 100% OK (that was my operation before Quickadd)
![Screen Shot 2021-06-24 at 15 18 06](https://user-images.githubusercontent.com/85180696/123317144-0aa36d00-d504-11eb-9f3b-20edea44b815.png)

Then I decided to investigate before writing to you. I did the following tests:
- Quickadd only with hotkey `Workbench: Copy the Current Line/Block to your Workbench` => 100% worked! 🟢
- Quickadd only with hotkey `Workbench: Link the Current Line/Block into your Workbench` => 100% worked!🟢
- Quickadd with 3 hotkeys `Workbench: Copy... + Show Outline Pane + Workbench: Link ... ` => fail 🔴  

[...]

Issue #33, opened 2021-06-25
[BUG] MACRO - Editing Obsidian Command


FelipeRearden:
Hello again @chhoumann !!!

The following MACRO has 5 Obsidian commands in this order: A - B - C - D - E 
![Screen Shot 2021-06-24 at 15 34 25](https://user-images.githubusercontent.com/85180696/123399019-86d89780-d57a-11eb-9e9b-422c3eec7595.png)

A few days later, I decided to change the order of the same MACRO to: B - C - D - E - A
![Screen Shot 2021-06-24 at 15 37 54](https://user-images.githubusercontent.com/85180696/123399059-8fc96900-d57a-11eb-8695-dcfa1dc76e66.png)

I close the window and when I come back, the order is again: A - B - C - D - E  
![Screen Shot 2021-06-24 at 15 35 03](https://user-images.githubusercontent.com/85180696/123399231-be474400-d57a-11eb-939f-8bf7d9268aec.png)

And the MACRO is respecting the A - B - C - D - E order when running.

I hope you could reproduce on you end. 
Have a great day!

chhoumann:
Thank you very much! I have addressed this in 0.2.11. Please confirm if this has been fixed for you.

FelipeRearden:
Hello @chhoumann!

I confirm that 0.2.12 fix the order issue. Now I can change the Obsidian commands position!!!!

The MACRO is amazing. I'm having a great time using it.

Thank you again @chhoumann !!!!! I wish you the best!

chhoumann:
I'm glad to hear that it's resolved, @FelipeRearden. Thank you, and I wish you the best, too. :)

PR #34, opened 2021-06-25
Quick Commands

chhoumann:


chhoumann:
This introduces Quick Commands. Ironically, we're starting with the Wait command.
The wait command should fix #32, and I've implemented fixes for #33.

PR #170, opened 2022-01-20
Make the text capture element multiline

### LINKED ISSUE #114: [FEATURE REQUEST] Capture Multi-line input
Similar to this https://twitter.com/TfTHacker/status/1439178293618618371

image

alexlyzhov:
This is a suggestion of a change to resolve #114. Seems to work for me but:

- IDK if sometimes single-line input may be preferable?
- This is the first time I interact with TypeScript and this repo so I may have missed something, please check.
- I didn't test it on mobile.
- The text area element should probably be wider, but in order to achieve this, ideally, the window itself needs to be wider. I could do smth like `textAreaComponent.inputEl.cols = 60;` but not sure if this is a one-size-fits-all solution.


chhoumann:
Hey @alexlyzhov! Thank you very much for spending time on contributing!

I actually already had a prototype for this, including how the different parts should interact, etc. etc.
There were a few things that I believed important to include, so I implemented a solution which seems to work on both desktop and mobile. I'll make a release with it now.

PR #220, opened 2022-03-25
citationsManager imports keywords as tags

Babakfifoo:
I am not a JS or TS developer, But I managed to add a few lines so the keywords of a citation can be imported as tags.
This might require an option to disable this feature if one does not need it.

Also in your guide in [here](https://bagerbach.com/blog/how-i-read-research-papers-with-obsidian-and-zotero#creating-notes-in-obsidian-from-papers-in-zotero) need to add the keywords to the template.

p.s. in the guide the tag for naming the new note should be {{VALUE:fileName}} while it is {{VALUE:title}} as causes issue with invalid characters.

### DIFF
--- a/docs/Examples/Attachments/citationsManager.js
+++ b/docs/Examples/Attachments/citationsManager.js
@@ -69,6 +69,7 @@ async function handleCitationsPlugin(params, citationsPlugin, settings) {
         zoteroSelectURI: entry.zoteroSelectURI,
         type: entry.type,
         issuedDate: entry.issuedDate,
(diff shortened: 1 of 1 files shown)


chhoumann:
Hey @Babakfifoo,

Thank you very much!
The file has been moved since your PR, so I've [updated it with you as a co-author](https://github.com/chhoumann/quickadd/commit/d23679e1e91ebc7d30a7e7461a74383c4a711d11).

Will update the guide as well!

PR #247, opened 2022-05-20
Explain QuickAdd more extensively in the README

martijnarts:
Hey! I love QuickAdd, but it's quite hard to figure out what's going on from just the README and the video. I ended up watching other Youtube videos on QuickAdd to learn what your plugin actually does.

While this explanation is probably not perfect, I think it's at least better than what was there before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,23 @@
 # QuickAdd
 Quickly add new pages or content to your vault.
+
+QuickAdd is a powerful combination of four tools (called choices): templates, captures, macros, and multis.
+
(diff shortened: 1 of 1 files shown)


chhoumann:
Hi @martijnarts 

Thank you very much for taking the time to write this.
Merged :)

martijnarts:
Hey, it does look like the merge duplicated some of the sections if you want to have a look at that. It doesn't seem to be a major problem, just the video is now shown twice.

chhoumann:
You're right, good catch! Thanks. Should be fixed now.

PR #346, opened 2022-12-12
Update MultiChoice.md

generic-user:
Making it clear in docs dragging and dropping is kinda hard ;) 

Thanks!


chhoumann (approved):
Hi @generic-user 

Thank you very much for clarifying this. Many users have mentioned it before, so I think it's good to get in the documentation.

Now the aim is to make the drag & drop experience even better! :)

PR #354, opened 2022-12-31
Prevent submit if press enter in composition session

kawamataryo:
Thanks for a very nice plugin!

I'm a Japanese user.
This plugin had a problem in the Capture input modal where the value being typed was being sent when a Japanese Kanji conversion was confirmed with the enter key. 
So I have fixed it.

CleanShot 2022-12-31 at 11 04 58

reference: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing


chhoumann (approved):
Hey @kawamataryo,

Thank you very much for the contribution!
Looks great.

Didn't even know about this, so thank you for teaching me something new :)

Issue #377, opened 2023-02-07
[FEATURE REQUEST] Access to {{VARIABLE:name}} in JavaScript blocks of captures.


silvio:
Maybe this is already possible and I just don't know.
It would be good to be able to access variables of the input mask in the javascrip block.

The background is that I use QuickAdd to collect sayings/quotes and then want to reference them. The quotes are stored in call-out boxes with a title. The title should also be the reference link. Here is my attempt:

    > [!quote] {{VALUE:title}}
    > {{VALUE:text}}
    > -- {{VALUE:author}}
    ```js quickadd
    const value = "{{VALUE:title}}";
    reformated = value.replace(/[^A-Z0-9]+/ig, "-");
    return `^${reformated}`;
    ```

The result should something like this:

    > [!quote] chaos nature order man
    > Chaos was the law of nature; Order was the dream of man.
    > -- Henry Adams
    ^chaos-nature-order-man

I tried several combinations with `${`/`${{`/`${name`/....

chhoumann:
Here's how you can do it:

````markdown
> [!quote] {{VALUE:title}}
> {{VALUE:text}}
> -- {{VALUE:author}}
```js quickadd
const title = await this.quickAddApi.inputPrompt("title");
reformated = title.replace(/[^A-Z0-9]+/ig, "-");
this.variables.title = `^${reformated}`;
```
````

Inline scripts are parsed and executed before anything else, so using `{{VALUE:title}} and then grabbing `this.variables.title`  isn't possible, as it won't have gotten the value yet. So instead, you'd ask for the title, do the replacement action, and then set the variable, which is then used above.

silvio:
Hi @chhoumann.

Thanks for your answer.

I wanted to use the title to create the block reference (last line, preceded by `^tile-with-hyphens` ) so that it is recognizably associated with the title.
Besides, this saves me a bit of extra manual typing and I don't have to ask about the reference.

Then this probably won't work with QuickAdd, will it?

chhoumann:
Hmm. When I try, it's only asking me for the title once. I imagine you're trying to use `{{VALUE:title}}` in a template?

If so, then a macro is suitable. You can use a macro to prompt for the input first, format it, and then you can use `{{VALUE:title}}` in the remaining steps.

E.g. have a script like this one in your vault `script.js`:
```js
module.exports = async function(params) {
    const title = await params.quickAddApi.inputPrompt('Enter a title');
[...]

PR #378, opened 2023-02-07
doc: inlinescript: Add sentence about parsing

silvio:
A bit more explanation why access to `{{VALUE}}` is not possible.

See #377.


chhoumann (approved):
Thank you, @silvio!

PR #390, opened 2023-02-19
BookFinder added to QuickAdd as an example

mlevison:
Christian as discussed last week on Discord. This is my first pull request ever.

For the code I've commented the obvious weaknesses:
- really should have a picker saying found X books with this name. Such a picker is made more complicated because some results from the Google Books api come back with undefined title (Bad API)
- Also I didn't take the time to elegantly concatenate all of the authors/categories of book with commas

I also wrote rudimentary docs, shamelessly stealing from your movie Script.


chhoumann (approved):
Hi @mlevison 

This is fantastic! Thank you very much for sharing.

PR #391, opened 2023-02-20
Add option to set default behavior if file already exists for template

cdloh:
Added an option to define the default behaviour if a file already exists at the creation path for a template.

If the behaviour is set then the user won't be prompted when the macro etc runs.

It's highly likely that the increment setting should be rolled into this but I wasn't entirely sure how to go about migrating settings etc


chhoumann (approved):
Hi @cdloh 

Thank you for contributing! This looks great.

Will figure out whether how to best deal with this & increment option based on user feedback.

PR #403, opened 2023-02-28
Added new format syntax {{FIELD:<variable>}}, and semi tab autocomplete

qaptoR:
FIELD searches all frontmatter for <variable> in the frontmatter of all markdown files.

Also implements partial autocomplete when pressing tab. Must have desired match under selection, otherwise it will just update to the current match (itself)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -2,6 +2,7 @@ export const VALUE_SYNTAX: string = "{{value}}";
 export const DATE_SYNTAX = "{{date}}";
 export const NAME_SYNTAX = "{{name}}";
 export const VARIABLE_SYNTAX = "{{value:<variable name>}}";
+export const FIELD_VAR_SYNTAX = "{{field:<field name>}}";
 export const MATH_VALUE_SYNTAX = "{{mvalue}}";
 export const LINKCURRENT_SYNTAX = "{{linkcurrent}}";
 export const TITLE_SYNTAX = "{{title}}";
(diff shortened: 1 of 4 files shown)


chhoumann (approved):
Hey! This looks good.

I've added handling for the case when there are no items found + added some error handling for `chooser`.

Thanks!

qaptoR:
@chhoumann 

Thank you for your feedback and support. tbh you're source code is very easy to read and work with.

I did notice some places in my build where I didn't account for everything as you had previously mentioned might happen (such as when showing the example output when writing format strings).

also, just moments ago I fixed a bug in my logic. I forgot that the field value is not always a string, so the suggester doesn't work if one of the field values is a number or otherwise not a string.

Now that the PR is complete I don't know how to push my fixes, but it is essentially just line 113 seen here
image

I just added the call of .toString()

chhoumann:
I'm glad to hear that, @qaptoR!

That's a good catch. Creating a new PR is the way to go in that case.
Otherwise, I can also fix it and add you as coauthor :)

qaptoR:
because it's such a small fix I don't mind in the slightest if you apply the fix.

I hope I have the chance to contribute more to this plugin. It's an amazing idea, and greatly improves the efficacy of my recordkeeping.

chhoumann:
Done!

That sounds great. Glad to hear the plugin is useful to you! Looking forward to future contributions :)

PR #408, opened 2023-03-06
Added Info Dialog as part of the API

Cassian-Andor:
Added a simple info dialog to be used within scripts instead of an ugly JS alert or similar. Existing dialogs didn't fill this purpose as they return a value or aren't really meant to be "informative" only.


chhoumann:
Hey @Cassian-Andor. This looks great.

I thought it would be an improvement to
1. Show the button on the right
2. Allow users to pass either an array of strings or a single string.

Which, by my commits, you can see I've added to this PR.
Obsidian_NccnI11FAQ

Otherwise, it looks great. Thanks for this!

PR #409, opened 2023-03-06
Respect the openFile option when running a template and a file already exists

cdloh:
Presently it always opens the file


chhoumann (approved):
Good catch. Thank you @cdloh!

PR #457, opened 2023-04-13
fixed URL for multi choices

pivic:
The multi choices URL isn't an URL; I've fixed it.


chhoumann:
Thank you, @pivic, good catch!

PR #460, opened 2023-04-15
UI: Improve clarity of the main Settings window

mikkovedru:
Before:
2023-04-15_Main-window_2-Before
After:
2023-04-15_Main-window_2-After

I think that the changes are self-evidently for the better.

There are a lot of UI changes that should imho be implemented to drastically improve the UI/UX aspect, both for the main Settings view as well as in almost every other window. 

But the problem is that:
1. I don't know if these changes are welcome and thus I don't want to work in vain.
2. Especially because these small changes took an enormous amount of effort from me. My JS/TS skills are not good and I hate JS with a passion anyway. :-)

Therefore if these changes are good and you would like me to continue, I could design the changes and explain them in new Issues, but I will leave the implementation to you/other people. How does this sound?

### DIFF
--- a/src/gui/choiceList/AddChoiceBox.svelte
+++ b/src/gui/choiceList/AddChoiceBox.svelte
@@ -52,5 +52,6 @@
         font-size: 16px;
         padding: 3px;
         border-radius: 3px;
+        appearance: auto;
     }
 </style>
--- a/src/gui/choiceList/ChoiceListItem.svelte
+++ b/src/gui/choiceList/ChoiceListItem.svelte
@@ -30,9 +30,10 @@
 	$: {
 		if (nameElement) {
 			nameElement.innerHTML = "";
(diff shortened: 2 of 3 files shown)


Dranaxel:
Please merge this 😇 being able to see quickadd types are a great improvement

chhoumann:
Hi @mikkovedru, thank you for your contribution here. I'm sorry I never had the time to properly review it back when we discussed it!

After review, I've decided to close this PR. There are a few reasons for this:
- Choice settings should be at the top. They are what you interact most with in settings, relative to the other, general settings.
- The styling should be Obsidian-native, and should account for a wide variety of themes and user preferences.

I've taken the feedback about unclear UX to heart over the past few years, and especially lately. The current settings panel is close to "fine for now," but I'm always open to suggestions or feedback.

Thanks again for taking your time to contribute!

PR #542, opened 2023-08-01
Update MacroChoice.md

danielo515:
there is a reference to an app variable that was never created


chhoumann (approved):
Thank you @danielo515!

Issue #545, opened 2023-08-03
[FEATURE REQUEST] "Create line if not found" at Top inserted AFTER frontmatter


n-konig:
# Description
When using "Insert after" -> "Create line if not found" with setting "Top", the line created displaces the frontmatter (as it really does create the line at the very top of the file!)

# Suggestion
Could the "Create line if not found" ignore the fronmatter, so that creating the line at the "Top" actually creates the line after the frontmatter is closed? This would be brilliant!

# Additional context
For example

```
---
key: value
---
```

now becomes

```
# Line created at the Top
---
key: value
---
```

but would preferably become

```
---
key: value
---
# Line created at the Top
```

Regards,
Niclas

f3d3r1c098:
is this issue back on version 1.8.1?

PR #595, opened 2023-10-27
Resolving Issue #545: "Create line if not found" at Top inserted AFTER frontmatter

evanaze:
## Description

This is a solution for issue #545 . As of Obsidian 1.4.0, the frontmatter position got moved from `frontmatter.position` to `frontmatterPosition`. See here: https://github.com/obsidianmd/obsidian-api/issues/122

As a result, the `getFrontmatterEndPosition` method in captureChoiceFormatter always returns `-1`, resulting in the note being added to the top of the file. This PR resolves that change in a backward compatible way by checking for both `frontmatter.position` and `frontmatterPosition`.

## Testing

Unfortunately, I don't know how to test my changes. Any tips here would be appreciated.

### DIFF
--- a/src/formatters/captureChoiceFormatter.ts
+++ b/src/formatters/captureChoiceFormatter.ts
@@ -25,7 +25,7 @@ export class CaptureChoiceFormatter extends CompleteFormatter {
 		input: string,
 		choice: ICaptureChoice,
 		fileContent: string,
-		file: TFile
+		file: TFile,
 	): Promise<string> {
 		this.choice = choice;
 		this.file = file;
@@ -36,7 +36,7 @@ export class CaptureChoiceFormatter extends CompleteFormatter {
 		const templaterFormatted = templaterParseTemplate(
 			this.app,
 			formatted,
-			this.file
+			this.file,
 		);
 		if (!(await templaterFormatted)) return formatted;
 
@@ -45,7 +45,7 @@ export class CaptureChoiceFormatter extends CompleteFormatter {
 
 	public async formatContent(
 		input: string,
-		choice: ICaptureChoice
+		choice: ICaptureChoice,
 	): Promise<string> {
 		this.choice = choice;
 		if (!choice) return input;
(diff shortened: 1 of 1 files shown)


src/formatters/captureChoiceFormatter.ts
```diff
@@ -171,8 +171,12 @@ export class CaptureChoiceFormatter extends CompleteFormatter {
...
 		}
 
-		if (fileCache.frontmatter.position)
-			return fileCache.frontmatter.position.end.line;
+		if (fileCache.frontmatter.position || fileCache.frontmatterPosition)
+			if (fileCache.frontmatter.position) {
+				return fileCache.frontmatter.position.end.line;
+			} else {
+				return fileCache.frontmatterPosition.end.line;
+			}
```
wadevries: Per suggestion in https://github.com/obsidianmd/obsidian-api/issues/122#issuecomment-1664346884. Also cleans up code a bit.
```suggestion
        const frontmatterPosition = fileCache.frontmatterPosition || fileCache.frontmatter?.position
        if (frontmatterPosition) {
            return frontmatterPosition.end.line;
```

evanaze: Thanks! Applied to the PR.

evanaze:
@chhoumann please review.

chhoumann (approved):
Thank you very much for the fix, @evanaze! This is perfect. I applied a small code style fix, but otherwise really nice.

chhoumann:
fixes #595

src/formatters/captureChoiceFormatter.ts
```diff
@@ -171,16 +170,19 @@ export class CaptureChoiceFormatter extends CompleteFormatter {
 			return -1;
 		}
 
-		if (fileCache.frontmatter.position)
-			return fileCache.frontmatter.position.end.line;
+		if (fileCache.frontmatter.position || fileCache.frontmatterPosition) {
+			if (fileCache.frontmatter.position) {
+				return fileCache.frontmatter.position.end.line;
+			}
+		}
```
edongashi, line 177: This change isn't actually doing anything because the `fileCache.frontmatterPosition` is discarded. @evanaze did you mean do discard `frontmatter.position` instead of `frontmatterPosition`?

PR #605, opened 2023-11-13
Fix typing error in suggester

mProjectsCode:
It seems like there was an error in the types of the `suggester` function. The functions seems to be used to map the elements to a string array, so the lambda should return a string, as it is called on every element in the array.


mProjectsCode:
https://quickadd.obsidian.guide/docs/QuickAddAPI#suggester-displayitems-string--value-string-index-number-arr-string--string-actualitems-string-promisestring

PR #694, opened 2024-05-31
include `replaceLinkToCurrentFileInString` in complete formatter

### LINKED ISSUE #400: [BUG] {{linkcurrent}} does not work within the Insert After field
When using Capture, and setting the `Insert After` and `Capture Format` fields with `{{linkcurrent}}`, it works fine inside Capture Format but does not work within Insert After. The result is that it inserts a literal text of {{linkcurrent}} instead of converting it. Again, the conversion works fine when using {{linkcurrent}} inside of Capture Format, it converts it to the original source filename as expected.

I need the ability to use the original filename as the Header that the text is captured into the new note.

QuickAdd v0.11.3

Screenshot below, the {{linkcurrent}} is the text `2023-02-20` which you see is converted properly in the text but not inside the header that QuickAdd creates. Note: You'll see the settings below shows QuickAdd actually detects the {{linkcurrent}} properly in its preview above the field, so I think this is a bug.

image

image
image

vsalavatov:
fixes #400

### DIFF
--- a/src/formatters/completeFormatter.ts
+++ b/src/formatters/completeFormatter.ts
@@ -45,7 +45,7 @@ export class CompleteFormatter extends Formatter {
 		output = await this.replaceVariableInString(output);
 		output = await this.replaceFieldVarInString(output);
 		output = await this.replaceMathValueInString(output);
-
+		output = await this.replaceLinkToCurrentFileInString(output);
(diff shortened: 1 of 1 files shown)


Morganwan90:
it works!! Thanks a lot!

PR #721, opened 2024-08-02
fix: heading pattern issue with leading tag

### LINKED ISSUE #720: [BUG] "Insert at end of section" doesn't recognize leading tag
**Describe the bug**
With "Insert after" and "Insert at end of section" option enabled, it's expected to insert new content at the end of `#Fleetings` section.

image

If there are a tag at the beginning of a line, quickadd will insert new content before the line of that tag, splitting the original content at that section.

It seems that quickadd recognizes it as a heading.

image

**To Reproduce**
1. enable "Insert after" and "Insert at end of section" option
2. write some content at the target section, with a tag independently at a line.
3. trigger the quickadd capture
4. new content is inserted before the tag, which is part of the section

**Expected behavior**
Parse the leading tag correctly, and insert new content at the actual end of section.

**Screenshots**
See above.

 - OS: macOS
 - Obsidian Version: 1.6.7 (Installer 1.5.11)

wrayzheng:
Fix #720.

In obsidian headings, there is at least one space after the hash sign (#), otherwise it should be treated as a tag.

### DIFF
--- a/src/formatters/helpers/getEndOfSection.test.ts
+++ b/src/formatters/helpers/getEndOfSection.test.ts
@@ -1,5 +1,5 @@
 import { test, expect } from "vitest";
-import getEndOfSection from "./getEndOfSection";
+import getEndOfSection, { getMarkdownHeadings } from "./getEndOfSection";
 
 test("getEndOfSection - find the end of a section", () => {
 	const lines = [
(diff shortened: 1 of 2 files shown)


chhoumann:
Hey @wrayzheng, thank you for this fantastic PR!

I added some additional test cases. And it made sense to me to change the optional `[\s]?` to match one or more spaces `[\s]+`, instead of adding it to the capture group `(#+\s)`. Otherwise we'd also capture whitespace while capturing the `#` symbols.
See: https://github.com/chhoumann/quickadd/pull/721/commits/99bc7ef08b0ad443d41f2aba197719ad0b6ed329

I really appreciate that you added tests. :)

PR #735, opened 2024-09-14
correcting log level

chaintng:
## Description

Found minor issue
`logWarning` function didn't call current log level

and might result in expected behaviour


chhoumann (approved):
Thank you, @chaintng!

PR #740, opened 2024-09-29
Update completeFormatter.ts to support arrays

WhiskeyJack96:
Updates the field validator to check if the value is of type array and if so adds each member of the array to the suggested items.

Tested in my vault with the `tags` field.

```yaml
tags: 
  - jrpg
  - game
  - foo
  - bar
```
will now properly render each option

### DIFF
--- a/src/formatters/completeFormatter.ts
+++ b/src/formatters/completeFormatter.ts
@@ -121,7 +121,7 @@ export class CompleteFormatter extends Formatter {
 			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 			const value = cache?.frontmatter?.[variableName];
 			if (!value) continue;
-			if (value.constructor === Array): value.forEach(x => suggestedValues.add(x.toString))
+			if (value.constructor === Array) value.forEach(x => suggestedValues.add(x.toString));
 			if (typeof value == "object") continue;
 
 			suggestedValues.add(value.toString());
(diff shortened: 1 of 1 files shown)


src/formatters/completeFormatter.ts
```diff
@@ -120,7 +120,9 @@ export class CompleteFormatter extends Formatter {
 			const cache = this.app.metadataCache.getFileCache(file);
 			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 			const value = cache?.frontmatter?.[variableName];
-			if (!value || typeof value == "object") continue;
+			if (!value) continue;
+			if (value.constructor === Array): value.forEach(x => suggestedValues.add(x.toString))
+			if (typeof value == "object") continue;
```
WhiskeyJack96:
```suggestion
			if (!value) continue;
			if (value.constructor === Array) value.forEach(x => suggestedValues.add(x.toString));
			if (typeof value == "object") continue;
```

src/formatters/completeFormatter.ts
```diff
@@ -120,7 +120,9 @@ export class CompleteFormatter extends Formatter {
 			const cache = this.app.metadataCache.getFileCache(file);
 			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 			const value = cache?.frontmatter?.[variableName];
-			if (!value || typeof value == "object") continue;
+			if (!value) continue;
+			if (value.constructor === Array) value.forEach(x => suggestedValues.add(x.toString));
```
WhiskeyJack96:
```suggestion
			if (value.constructor === Array) {
			    value.forEach(x => suggestedValues.add(x.toString));
			    continue
			}
```

PR #750, opened 2024-11-03
fix: use frontmatterPosition end line

### LINKED ISSUE #647: [FEATURE REQUEST] Appending to existing file while considering the frontmatter
For now, If I use QuickAdd to append a existing note which frontmatter with this option:
image

It would just insert the text BEFORE the original frontmatter.. which would break the frontmatter function.

So I was wondering could there be a new option "Append to the top (Ignore frontmatter)"
And with this, put the content before the CONENT but AFTER the frontmatter ?
Thus not breaking thr original frontmatter.

edongashi:


chhoumann:
Hey @edongashi, thank you for taking the time to contribute!

It looks like the API may have changed since this was originally merged, IIRC. Glad you wrote a fix!

Your PR doesn't resolve #647 because that regards Templates, whereas this modifies code for Captures. So I removed the `fix` from your PR body to prevent that issue from closing when I merge this.

chhoumann (approved):

edongashi:
Thank you @chhoumann!
 
> It looks like the API may have changed since this was originally merged, IIRC. Glad you wrote a fix!

The API change was acknowledged in the previous PR #595, but I think by accident the wrong line was deleted [here](https://github.com/chhoumann/quickadd/pull/595#discussion_r1439247067)

> Your PR doesn't resolve https://github.com/chhoumann/quickadd/issues/647 because that regards Templates, whereas this modifies code for Captures.

Thanks for being vigilant, I tried to do some housekeeping given the large number of issues. I gave the issue a second look, and I don't see how this related to templates. Can you please double check?

chhoumann:
Of course. The contents of the issue sounds more relevant to a Capture, but specifically mentions a setting that is only available in Templates:
image

I checked & can recreate the issue they presented with the setting.
You would have to set the file name to be the same as the file name of an existing file. Then it would try to create the note, notice there's already a file, and instead proceed by simply adding the text to the top of the note, disregarding the frontmatter.

edongashi:
Understood, thanks and keep up the good work!

chhoumann:
Thank you, likewise!

Issue #583, opened 2023-10-10
[FEATURE REQUEST] Replacing the current tab with note after capturing to note


devinat1:
**Is your feature request related to a problem? Please describe.**
Whenever I capture to my daily note, it opens the note in a new tab (even if I don't have the new tab option turned off).

**Describe the solution you'd like**
I would like for the file I create to open in the current tab (replacing the current tab).

**Additional context**
Currently, every time I have to do ctrl+tab to go to the daily note after capturing to it, which is not ideal.

Issue #530, opened 2023-07-05
[FEATURE REQUEST] Focus on already open note


android272:
# Is your feature request related to a problem? Please describe.

Twice a day I run a capture command that adds some text to the bottom of a file and opens it so that I can modify the output a bit. If I leave the note open by close the app or what ever, when I go to run the command again 6 hours later or the next day a new tab will be opened for this note. If I am not concise about this on my phone where tabs are a bit more hidden I can easily have many tabs for the same note opened at the same time. 

# Describe the solution you'd like:
I would like the ability to switch to an already open tab if the file exists. This could be a how the `open setting` works by default or perhaps this is a bug because I have `new tab` setting turned off.

tauren:
+1 

I struggle with tab proliferation and 90% of it is because QuickAdd doesn't have a way to switch to an existing tab. I don't want to replace the current tab, but I do want to open the newly changed file. If that file is already open, I want to switch to it's tab, otherwise open a new tab.

I think #401 is raising the same issue.

JayPritchet:
I created a new pr #758 , it will check if there is an exsiting tab that already opened the file. If true, it will switch to that tab, otherewise it will created a new tab and focus on it.

Issue #401, opened 2023-02-25
[FEATURE REQUEST] Open capture in existing leaf / tab


jonathaneckman:
A capture with "Open" enabled always opens the file in a new leaf tab. This results in tab proliferation. 

I would like the ability to focus the existing leaf if it is already open or, if it is not open, open it then focus it. 

I tried to accomplish this using:

1. [No dupe leaves](https://github.com/scambier/obsidian-no-dupe-leaves) works for links outside of QuickAdd commands, but has no effect on the files opened from QuickAdd.
2. [Quick Switcher++](https://github.com/darlal/obsidian-switcher-plus) works when you invoke its commands from the command palette, but cannot be run from a QuickAdd macro because it requires an input parameter (the file you want to open), which cannot be provided through the QuickAdd configuration.

seomwan:
Duplicate of #338

I'd considered this as a regression because it used to work in Obsidian 0.5 before the introduction of tabs.
It's the same problem as yours, tab proliferation.

The current (best so far) workaround I have settled with is to not open any files for my most commonly used actions

Issue #338, opened 2022-11-28
[BUG] (Capture) choice doesn't open in active tab


seomwan:
**Describe the bug**
(Capture) choice doesn't open in the same tab as the current active when option not to open in new tab is set.
Context: 
> I'd prefer it to work that way as new tabs introduce clutter, especially on mobile where it's harder to close many tabs at once.

**To Reproduce**

Minimal sample config: [data.zip](https://github.com/chhoumann/quickadd/files/10107490/data.zip)

OR:

1. In current or new vault, create a capture choice that opens in the same tab:
  a. Enable `Open`
  b. Disable `New tab`
2. Run action (should open in new tab)
3. Run action again to confirm behavior (should open in new tab)

**Expected behavior**
Opens the file that is captured to in the current active tab

**Screenshots**
![image](https://user-images.githubusercontent.com/68874610/204379252-64e4732a-5224-4da1-97fd-8044e7b21f28.png)

**Desktop (please complete the following information):**
 - OS: Windows

**Smartphone (please complete the following information):**
 - Device: Samsung
 - OS: Android

**Additional context**
> I'd prefer it to work that way as new tabs introduce clutter, especially on mobile where it's harder to close many tabs at once.

Originally [posted here](https://github.com/chhoumann/quickadd/issues/290#issuecomment-1329728704)

PR #751, opened 2024-11-03
fix: prevent creating new tab when openInNewTab is not set

### LINKED ISSUE #634: [FEATURE REQUEST] Templates choice - a setting that opens the file created in the current tab
Current settings in **template choice** offer

- open the created file (yes/no)
- if set to open
  - which view mode is used (default/source/preview)
  - is vertical or horizontal split used
  - is the tab focused (yes/no), using this doesn't overwrite the global Obsidian setting\
  `Editor-> Always focus new tabs`

I propose new setting in **template choice** to _open in current tab_.

Similar currently open feature requests regarding **capture choice**:
- https://github.com/chhoumann/quickadd/issues/338
- https://github.com/chhoumann/quickadd/issues/401
- https://github.com/chhoumann/quickadd/issues/530 (the title is inaccurate)
- https://github.com/chhoumann/quickadd/issues/583

edongashi:
Fixes #634, #583, #530, #401, #338.

As per API docs:

> If newLeaf is false (or not set) then an existing leaf which can be navigated
> is returned, or a new leaf will be created if there was no leaf available.

Note that settings must be enhanced to support an additional state (option 3):

1. Current tab (when open in new tab is unchecked) - fixed in this PR
2. New tab with split (vertical/horizontal) - exists
3. New tab without split - handled if `direction` is not set. Current settings do not support this.

I guess most people would like 1 and 3. 2 is a bit niche, as splits are annoying.

### DIFF
--- a/src/utilityObsidian.ts
+++ b/src/utilityObsidian.ts
@@ -139,8 +139,10 @@ export async function openFile(
 
 	if (optional.openInNewTab && optional.direction) {
 		leaf = app.workspace.getLeaf("split", optional.direction);
-	} else {
+	} else if (optional.openInNewTab) {
 		leaf = app.workspace.getLeaf("tab");
+	} else {
+		leaf = app.workspace.getLeaf(false);
 	}
 
 	await leaf.openFile(file);


JayPritchet:
Hi, It seems the changes in your commit will not achieve the goal.

I tested your code, it will close the opened file in current active tab, and then open the desired file in the same tab.

I created a new pr, it will check if there is an exsiting tab that already opened the file. If true, it will switch to that tab, otherewise it will created a new tab and focus on it.

chhoumann:
Hi @edongashi, thank you for taking your time to contribute. I've merged a modified version of @JayPritchet's code and will therefore close this PR.

PR #758, opened 2024-11-26
fix: prevent creating new tab when file is already opened

### LINKED ISSUE #634: [FEATURE REQUEST] Templates choice - a setting that opens the file created in the current tab

JayPritchet:
Fixes #634 #583 #401 #338 #530 
Added a new setting item "Focus Existing Tab" in temple and capture setting.

When "Focus Existing Tab" is set True, the obsidian will check if there is a tab that already opened the file. If already opened, it will set that tab to be actived, otherewise it will create a new tab with corresponding settings.

When "Focus Existing Tab" is set False, the obsidian create a new tab with corresponding settings.

The new setting panel is like this:
image

### DIFF
--- a/src/engine/CaptureChoiceEngine.ts
+++ b/src/engine/CaptureChoiceEngine.ts
@@ -90,14 +90,12 @@ export class CaptureChoiceEngine extends QuickAddChoiceEngine {
 			}
 
 			if (this.choice.openFile && file) {
-				let openExistingTab=false;
-				if(this.choice.focusExsitingFileTab)
-				{
-					openExistingTab=await openExistingFileTab(this.app, file);
+				let openExistingTab = false;
+				if (this.choice.focusExistingFileTab) {
+					openExistingTab = await openExistingFileTab(this.app, file);
 				}
 
-				if(!openExistingTab)
-				{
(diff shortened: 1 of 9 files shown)


chhoumann (approved):
Hey @JayPritchet, this is great! Thanks for the PR. I've made some modifications:

This is now the default behavior when opening. I think it's simpler than adding another option to the configurations. But if anyone wants it, I'll include it later.
I've also slightly modified the implementation due to some cross-compatibility concerns & such, but it's been good overall.

PR #849, opened 2025-07-06
Add InputSuggester to api

sualk01:
Adds the inputSuggester which opens a modal to select an item, similar to genericSuggester but with the possibility of also creating an item.

I'm not sure if the options parameter has the right type assigned to it in quickAddApi.tx


sualk01:
already implemented with 2.0

sualk01:
No worries, thanks for implementing it!

PR #858, opened 2025-07-15
Remove macro indirection - integrate macro management directly into choices

### LINKED ISSUE #543: [FEATURE REQUEST] Less macro indirection
Hello, and thank you for this great plugin.
This more than a feature request is a question about why macros require so much indirection.
It seems not possible to select several macros in the select macro option, so I don't understand  why a separate step is needed. Why have a button to manage macros, and another button to create a new macro choice if all you can do in the macro choice is say which macro to execute? Why not just allow to configure the macro directly from the macro choice?
The only use case I can think of, is if you want to alias your macros, but that's all

### LINKED ISSUE #816: Add unused macros detection feature
## Summary
Add functionality to detect and highlight macros that are no longer used or referenced anywhere in QuickAdd configurations.

## Problem
Over time, users accumulate many macros and may:
- Create test macros that are never deleted
- Remove references to macros but forget to delete the macro itself
- Have orphaned macros from changed workflows
- Struggle to identify which macros are actually in use

## Proposed Solution
Add **"Unused Macros Detector"** feature in QuickAdd settings:

### Detection Logic
- Scan all Choice configurations for macro references
- Check for macros called via API in user scripts
- Identify macros never referenced anywhere
- Show last-used dates if available

### UI Features
- **"Find Unused Macros"** button in settings
- List view showing potentially unused macros
- Confirmation before bulk deletion
- Export unused macros before deletion (for backup)

## Implementation Ideas
- Scan choice configurations for macro calls
- Parse user scripts for `quickAddApi.executeChoice()` calls
- Track macro usage statistics over time
- Safe deletion with confirmation dialogs

## Benefits
- Cleaner macro organization
- Reduced cognitive load when browsing macros  
- Better performance with fewer unused macros
- Easier maintenance of large macro collections

## Related
Similar to "unused CSS" detectors in web development tools.

chhoumann:
This PR implements the feature request from #543 by eliminating the confusing two-step process where users had to manage macros separately then create macro choices that referenced them.

### User Experience Improvements
- **Before**: User creates macro choice → MacroChoiceBuilder opens → User clicks 'Configure Macro' → MacroBuilder opens
- **After**: User creates macro choice → MacroBuilder opens directly with integrated choice settings

### Technical Implementation
- **Eliminated MacroChoiceBuilder entirely** - No more intermediate step
- **MacroChoice now embeds IMacro directly** - True 1:1 relationship without indirection
- **Moved runOnStartup from IMacro to MacroChoice** - Logical placement on the choice level
- **Enhanced MacroBuilder** - Now handles entire MacroChoice (name + runOnStartup + macro)
- **Perfect name synchronization** - Choice name and macro name stay in sync automatically
- **Comprehensive migration system** - All existing data is preserved and converted seamlessly

### Files Changed
- **Removed**: `MacrosManager.ts`, `MacroChoiceBuilder.ts`
- **Enhanced**: `MacroBuilder.ts` - Now handles full MacroChoice management
- **Updated**: All engines, formatters, and settings to use embedded macros
- **Added**: `removeMacroIndirection.ts` migration for backward compatibility

### Testing
- ✅ All existing tests pass
- ✅ TypeScript compilation successful
- ✅ Build completes without errors
- ✅ Migration system tested with backward compatibility

## Migration Strategy
The migration system safely converts existing macro configurations:
1. Finds all macro choices that reference macros by ID
2. Embeds the referenced macro directly into the choice
3. Transfers runOnStartup setting from macro to choice
4. Creates new macro choices for any orphaned macros
5. Removes the global macros array

This ensures no data is lost and all existing functionality is preserved.

Resolves #543
Closes #816


chhoumann:
Also closes https://github.com/chhoumann/quickadd/issues/816 which is now redundant

Issue #847, opened 2025-06-30
[BUG] In prompt when choose tag from hint, double hashtag appears


Don-Touch:
**Describe the bug**
In prompt when choose tag from hint, double hashtag appears

**To Reproduce**
1. Create a QuickAdd template with prompt
2. In prompt type # and choose any of tags

**Actual behavior**
Two hashtags appears before the tag

**Expected behavior**
Only one hashtag expected

**Screenshots**

![Image](https://github.com/user-attachments/assets/69fb675f-9bd9-4149-a950-ebb4bf8a390b)

![Image](https://github.com/user-attachments/assets/7720c404-9a07-43b2-8666-c19251e241a5)

PR #860, opened 2025-07-24
fix: resolve tag selection truncation in text input fields (#859)

### LINKED ISSUE #859: [BUG] Tag selection doesn't work in text input field
**Describe the bug**
After #847 was fixed, now instead of a double hash, the text you typed to sloppy search a tag is used at the tag, instead of the full tag you select. This is very easy to find with sub tags (as in \#tag_category/sub_tag), but happens with any longer tag that is sloppy-matched. If multiple tags are selected, only the last tag is truncated to the sloppy search characters. This is happening when I select the full tag with the mouse or hitting enter. The full, completed tag is show in the text input field.

**To Reproduce**
Steps to reproduce the behavior in Obsidian:
1. For instance, sloppy search a tag in the form \#category/subtag with `#subtag`
2. Click Ok
3. See \#subtag instead of \#category/subtag in the resulting note.

**Expected behavior**
The selected tag should make it without change into the note.

**Screenshots**
I can produce screenshots if needed.

**Environment (please complete the following information):**
 - Obsidian version:  1.8.10
 - QuickAdd version: 1.18.1
 - Operating system: macOS Sequoia 15.5
 - Debug info: Paste output from Developer Tools (Ctrl+Shift+I or Cmd+Option+I on Mac)
 - Other plugins enabled: TagWrangler 0.6.4, many others but nothing elseseems like it would be related at all
 - Theme: Minimal

**Console Logs**
If applicable, open developer tools (Ctrl+Shift+I or Cmd+Option+I on Mac) and paste any relevant error messages.

**Mobile (if applicable, complete the following information):** I don't use QuickAdd for this on Mobile
 - Device: [e.g. iPhone 12]
 - OS: [e.g. iOS 15]
 - Obsidian version: [e.g. 1.0.0]
 - QuickAdd version: [e.g. 0.5.0]

**Additional context**

chhoumann:
## Description

Fixes issue #859 where tag selection in text input fields was truncating tags to just the search characters instead of using the full selected tag.

## Root Cause Analysis

The issue was introduced as a side effect of fixing #847 (double hashtag bug). The fix for #847 added  to prevent the suggestion popup from reappearing after tag completion. However, this suppressed the  event that  (Obsidian's wrapper) needs to synchronize its internal state with the DOM.

### The Problem Flow:
1. User types `#subtag` to search for `#category/subtag`
2. User selects the full tag from suggestions
3. `TagSuggester.selectSuggestion()` calls `replaceRange()` with `dispatchInput: false`
4. DOM shows `#category/subtag` but `TextComponent` internal state still has `#subtag`
5. When field blurs/re-renders, `TextComponent` writes its stale value back to DOM
6. Result: tag appears truncated to `#subtag`

## Solution

Following the same pattern used in `FileSuggester`, this fix:

1. **Keeps** `dispatchInput: false` to maintain the fix for #847
2. **Adds** manual `input.trigger("input")` to update `TextComponent` internal state
3. **Preserves** the correct execution order: replace text → trigger event → close suggester

## Code Changes

```typescript
// Before
replaceRange(input, hashPos, cursor, replacement, { dispatchInput: false });
this.close();

// After  
replaceRange(input, hashPos, cursor, replacement, { dispatchInput: false });
input.trigger("input");
this.close();
```

## Testing

This fix addresses the specific scenarios mentioned in #859:
- ✅ Sloppy search for hierarchical tags (`#category/subtag`) now inserts the full tag
- ✅ Multiple tag selections work correctly (last tag no longer truncated)
- ✅ Both mouse click and keyboard selection (Enter) work properly
- ✅ Maintains fix for #847 (no double hashtag)

## Compatibility

- No breaking changes
- Consistent with existing `FileSuggester` implementation
- Preserves all existing functionality

Fixes #859

### DIFF
--- a/src/services/choiceService.ts
+++ b/src/services/choiceService.ts
@@ -19,9 +19,6 @@ import { TemplateChoice } from "../types/choices/TemplateChoice";
 import { excludeKeys } from "../utilityObsidian";
 import { regenerateIds } from "../utils/macroUtils";
 
-/**
- * Constructor map is colocated here to keep service self-contained.
- */
 const choiceConstructors: Record<ChoiceType, new (name: string) => IChoice> = {
 	Template: TemplateChoice,
 	Capture: CaptureChoice,


chhoumann:
Waiting with this one. Doesn't seem to provide much of a difference in behavior

chhoumann:
Want to double check this but fixes the issue for me

PR #891, opened 2025-08-19
fix: adjust mobile styles for multi-line input prompt and clean up me…

mehanoid:
### Motivation

When the "Use Multi-line Input Prompt" option was enabled, the input dialog was displayed incorrectly:

- On mobile devices, the intended styles were overridden by the faulty “tablet” media query. The input happened to take the full width only by accident, but the mobile-specific layout was broken. The input field had too large a height and was positioned too low, so it slid under the on-screen keyboard.
- On tablets, styles did not apply at all due to the incorrect media query::

```
@media screen and (max-width: 540px) and (max-width: 780px) { ... }
```

As a result, the dialog looked broken.

In addition, the tablet-specific block only differed from desktop by setting a fixed input width, which was unnecessary and actually degraded the layout.

### Changes

- Fixed the mobile media query so that mobile-specific styles are now correctly applied.
- On mobile devices:
  - the modal is shifted above the keyboard (bottom: 10vh),
  - the input field height was increased (6rem), while keeping the correct width behavior.
- Removed the tablet-specific block: tablets now use desktop styles, which simplifies the layout and improves usability (input takes full width).

### Screenshots

Before (mobile):  

After (mobile):  

Before (tablet):  

After (tablet):

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -39,6 +39,10 @@
 
 /* Mobile */
 @media screen and (max-width: 540px) {
+    .qaWideInputPrompt .modal {
+        bottom: 10vh;
+    }
+
 	.macroContainer1 {
 		grid-template-columns: repeat(1, 1fr);
 	}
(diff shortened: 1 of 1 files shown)


chhoumann:
Hi @mehanoid, thank you for this PR! I think it looks great. Thank you for taking the time to contribute.

Issue #294, opened 2022-10-16
Misleading error message


talenelat-elin:
When I try to add a template with the Filename Format containing a `:` (or, I suspect, any other character that's not valid in a file name like `<>,?`), it throws the following error message, pointing towards a problem with the Template Path rather than the Filename Format.

![image](https://user-images.githubusercontent.com/71895835/196032234-236d96bc-3a1f-4adc-b52e-fac0a66d29b5.png)

I spent a couple of hours browsing through the github issues, to no avail. Then I accidentally discovered where the problem was, or I'd have stopped using this plugin altogether.

I've seen a couple of other reports mentioning that they have a valid template path but they're still getting this message, I'll link them later when I get the time. 

Suggestion: If the error message cannot be made more specific, could there be an additional clause in it that the error could be due to something else entirely?

I understand it says that "Maybe" the template path is the issue, but if it was made more clearer, I suspect it would help other users too.

Also seems related to this comment in #111 
>> > > Hi, sorry for commenting on a closed ticket, but this particular error-message also appears when using non-legal characters as a title/filename, such as a questionmark (?) on Windows-machines.
>> > Would you rather see a separate ticket for this?
>> 
>> It's all right. If an illegal character is given, then it can't create the file - so it's working as intended. Or am I missing something?
>In a sense, yes. But the problem is that the error-message is misleading as it points to an error with the template instead of the filename of the new note.

_Originally posted by @flatrick in https://github.com/chhoumann/quickadd/issues/111#issuecomment-981922054_

firmart:
I had the same issue and couldn’t understand from the error message what had gone wrong. The message said that File name cannot contain any of the following characters \/, but in fact it was the : that caused the problem.

chhoumann:
Hey! I agree this isn't ideal. I've pushed a fix which will be included in the next update.

firmart:
Thanks for the quick fix! Much appreciated.

chhoumann:
Hi @bennyyip - fix incoming! That was an unintended side-effect of the additional validation.

chhoumann:
Fix released. Please update to 2.2.1 and let me know if it works for you :) @bennyyip 
Works for me with your example.

PR #915, opened 2025-09-07
fix(validation): allow slashes in filename formats; remove pre-validation

### LINKED ISSUE #914: [BUG] Can't select File as psrt of Capture flow.
Before QuickAdd 2.2.0 - you could keep File blank then select the File as part of the Capture sequence. 

For an example, I could use the Add Idea capture to select a note, then append a line to that note. This was a very powerful and dynamic workflow. 

Now, QuickAdd 2.2.0 the Capture says "Error running capture choice. Invalid filename format. File name cannot be empty"

To reproduce, set blank Filename format. 

The desired behaviour is to have the pre 2.2.0 behaviour where you would have a 2 step process where you can set the note as part of the Capture flow.

chhoumann:
This PR fully reverts the restrictive filename validation introduced after #294 and restores the previous (and widely used) behavior where slashes in "File Name Format" are allowed to create subfolders.

Summary
- Remove proactive filename pre-validation (both Template and Capture flows)
- Allow slashes in file name formats again so patterns like `{{DATE:YYYY}}/W{{DATE:WW}}/{{DATE:}}` work as before
- Rely on platform-native errors for invalid names or paths, which provides clearer, OS-specific messages
- Improve error context when template file creation fails by including the target path

Rationale
- Users frequently rely on time-based subfolders in their file name formats (e.g., `YYYY/MM/<title>`)
- The strict pre-validation broke existing workflows and provided less actionable errors than native OS messages

Implementation Notes
- Removed filename validation utilities and their tests
- Removed validation calls from `TemplateChoiceEngine` and `CaptureChoiceEngine`
- Adjusted `TemplateEngine` error handling to surface the original platform error (with path context)

Closes #914


chhoumann:
In general, this blocked a lot of super useful workflows that people were relying on. The validation was way too harsh and it blocked some features that were part of QuickAdd for many years. For that reason, I've instead opted to just handle thrown errors from the platform that essentially highlights what exactly is allowed or not when creating a file that potentially has an invalid file name.

PR #924, opened 2025-09-18
feat: add optional append link modes

chhoumann:
## Summary
- add optional and required link-to-note modes for capture and template choices
- make link insertions resilient when no file is active
- document the new behavior for append link and {{LINKCURRENT}}

## Testing
- npm test


chhoumann:
Nice improvement! Could you add a short note to the release notes once merged so users notice the new optional mode?

PR #1275, opened 2026-06-01
fix: preserve prompt drafts until choice execution commits

### LINKED ISSUE #1157: [FEATURE REQUEST] Improvement: Store Capture if Note Creation fails
When trying to create a new note which contains an invalid character (in my case a ':'), a notification is shown that the note could not be created. However, the actual capture is lost, which can be very frustrating.

Different ideas come to mind: either a warning could be shown in the capture dialog, stating that the input cannot be captured since it contains invalid characters. Alternatively, the input could be stored in the users clipboard whenever file creation fails. This would probably be the more robust solution, since it also handles FS issues etc.

Simply not inputting invalid characters is of course an alternative, but that would mean that every user has to stub their toe on this problem at least once.

**Additional context**
I am happy to implement a PR for this if your workload is to high - please indicate if that's the case.

chhoumann:
Preserve submitted input prompt drafts until the surrounding choice execution has either committed or failed. This replaces the clipboard-primary recovery added in #1270 with the existing session-only `InputPromptDraftStore` path, and removes the eager clipboard overwrite on capture target creation failures.

The change adds execution scopes to `InputPromptDraftStore`, routes submitted prompt persistence through those scopes, and wraps `ChoiceExecutor.execute()` so successful choices clear submitted drafts while failed or aborted executions leave them available for retry. Capture and Template errors that are reported without throwing now mark the active draft scope failed so the executor does not commit those submitted prompt values.

Reproduction evidence before the fix: the new capture regression failed on `origin/master` with `store.get(...)` returning `undefined` after a submitted `{{VALUE}}` prompt and failed target-file creation.

Validation:
- `bun run test src/engine/CaptureChoiceEngine.selection.test.ts` failed before implementation with the submitted draft cleared.
- `bun run test src/engine/CaptureChoiceEngine.template-property-types.test.ts`
- `bun run test src/utils/InputPromptDraftStore.test.ts src/engine/CaptureChoiceEngine.selection.test.ts`
- `bun run test src/engine/TemplateChoiceEngine.notice.test.ts src/utils/InputPromptDraftStore.test.ts src/engine/CaptureChoiceEngine.selection.test.ts`
- `bun run test`
- `bun run build-with-lint`
- `bun run test:e2e tests/e2e/scorecard-composed-flows.test.ts` was attempted; it failed in e2e setup before assertions while waiting for the sandbox template file content to be observed by Obsidian, and the generated `__obsidian_e2e__/` artifact was removed.

Obsidian dev-vault validation:
- Built the PR bundle with `bun run build`.
- Temporarily pointed the dev-vault QuickAdd plugin symlinks at the PR worktree bundle, reloaded with `obsidian vault=dev plugin:reload id=quickadd`, then restored the symlinks to `/Users/christian/Developer/quickadd/...`.
- Used `obsidian vault=dev dev:debug on`, `dev:console clear`, `dev:errors clear`, and detached with `obsidian vault=dev dev:debug off`.
- Created and removed temp choice `__qa-pr1275-draft-capture` via `data.json` backup restore.
- Verified failed capture target creation preserves the submitted VALUE draft and does not overwrite the clipboard.
- Verified retry hydrates the submitted draft, a same-key retry can succeed after a one-time `app.vault.create` failure, and the next prompt opens empty after success, confirming the submitted draft clears on commit.
- `obsidian vault=dev dev:errors` reported no captured runtime errors after validation.
- Cleaned temp file `__qa-pr1275-success.md`; no PR1275 temp leftovers found in the dev vault. Dev-vault slot is released.

Refs #1157
Refs #1270

## Summary by CodeRabbit

* **Tests**
  * Added execution-scope tests to verify prompt-draft persistence, rollback, commit, failure and reuse behaviors across runs.

* **Bug Fixes**
  * Stop silently falling back to copying capture content to the clipboard after failures.
  * Preserve submitted prompt drafts across non-abort failures, reuse them on retry, and clear them only on successful commit or explicit clear.

* **Refactor**
  * Improved execution-scope management so in-progress and submitted prompt drafts behave consistently during multi-step flows.

### DIFF
--- a/src/choiceExecutor.ts
+++ b/src/choiceExecutor.ts
@@ -15,6 +15,7 @@ import { runOnePagePreflight } from "./preflight/runOnePagePreflight";
 import { MacroAbortError } from "./errors/MacroAbortError";
 import { isCancellationError } from "./utils/errorUtils";
 import { getOpenFileOriginLeaf } from "./utilityObsidian";
+import { InputPromptDraftStore } from "./utils/InputPromptDraftStore";
 
 export class ChoiceExecutor implements IChoiceExecutor {
 	public variables: Map<string, unknown> = new Map<string, unknown>();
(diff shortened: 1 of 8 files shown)


chhoumann:
@coderabbitai review

PR #1276, opened 2026-06-01
feat: option to insert template or capture links into the frontmatter of active note

ArjanSeijs:
## Summary

This adds an option to insert links for templates and captures into the frontmatter of the active note. For my TTRPG vault i make heavy use of quickadd and needed this feature to quickly add links to the frontmatter of a file. For example for quickly adding npcs or sub locations to the frontmatter of my towns.

## Changes

* option to insert template or capture links into the frontmatter of active note

## Testing / validation
Obsidian version 1.12.7
I updated the tests in `linkPlacement.test.ts` to account for the new LinkPlacement. 
Testing the feature was done manually, see the gif for results.

Screenshot_2026-06-01_21-15-06

output

## Checklist

- [x] Linked any related issue(s).
- [x] Noted release/migration impact, if any.

---

<a href="https://app.devin.ai/review/chhoumann/quickadd/pull/1276" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added **“In frontmatter property”** as a link placement option for capture and template choices, including controls to choose the target property and appending behavior.
  * Supports inserting link text into frontmatter when the active file is available.

* **Documentation**
  * Updated choice docs to explain how behavior works when the targeted frontmatter property is not a list (including create/append/error handling).

* **Bug Fixes / Improvements**
  * Ensures capture and template link insertion completes before subsequent actions continue.

* **Tests**
  * Added coverage for frontmatter insertion behavior and placement support.

### DIFF
--- a/docs/docs/Choices/CaptureChoice.md
+++ b/docs/docs/Choices/CaptureChoice.md
@@ -59,6 +59,7 @@ If you have a tag called `#people`, and you type `#people` in the _Capture To_ f
     -   **After selection** - Preserves selected text and places the link after it
     -   **End of line** - Places the link at the end of the current line
     -   **New line** - Places the link on a new line below the cursor
+    -   **In frontmatter property** - Adds the link to a frontmatter property
 
 ## Canvas Capture Notes
 
--- a/docs/docs/Choices/CaptureChoice.md
+++ b/docs/docs/Choices/CaptureChoice.md
@@ -61,6 +61,11 @@ If you have a tag called `#people`, and you type `#people` in the _Capture To_ f
     -   **New line** - Places the link on a new line below the cursor
     -   **In frontmatter property** - Adds the link to a frontmatter property
 
+        When _In frontmatter property_ is selected you can choose what to do when encountering non-list properties:
+        - **Throw an error** -  Throws error if the property is not a list or does not exist. Empty properties are handled as an empty list.
+        - **Create property** - Adds the property to frontmatter if it does not exist. Throws an error if the property already exists and is not a list. Empty properties are handled as an empty list.
+        - **Always append** - Will convert non-list properties to a list and append the value.
+
 ## Canvas Capture Notes
 
 QuickAdd supports two Canvas capture workflows:
--- a/docs/docs/Choices/TemplateChoice.md
+++ b/docs/docs/Choices/TemplateChoice.md
@@ -31,6 +31,10 @@ When either enabled mode is selected, you can choose where the link is placed:
 - **New line** - Places the link on a new line below the cursor
 - **In frontmatter property** - Adds the link to a frontmatter property
 
+    When _In frontmatter property_ is selected you can choose what to do when encountering non-list properties:
+   - **Throw an error** -  Throws error if the property is not a list or does not exist. Empty properties are handled as an empty list.
+   - **Create property** - Adds the property to frontmatter if it does not exist. Throws an error if the property already exists and is not a list. Empty properties are handled as an empty list.
+   - **Always append** - Will convert non-list properties to a list and append the value.
 
 **If the target file already exists**. Choose whether QuickAdd should ask what to do, update the existing file, create another file, or keep the existing file.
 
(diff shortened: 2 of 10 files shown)


ArjanSeijs:
I can add that that the frontmatterProperty persists after changing the settings, however this not the case for the linkType since this is also just reset when switching between require vs skip active file.

For the processFrontMatter
1. I can make it async but this would require to make a lot of functions async, 
2. or would you rather catch the error and create a new notice to inform the user.

chhoumann:
Hey!

Thanks for working on this. I like the feature direction, and I can see why it would be useful for TTRPG/worldbuilding vaults.

I did some local testing in the dev vault, and I think this needs a bit more work before I can merge it.

On `processFrontMatter`: I’d prefer making the call chain async and awaiting it, rather than catching locally and showing a notice. The reason is that QuickAdd should know whether the link insertion actually succeeded. Right now the command can continue/report success while the frontmatter write fails later. In my dev-vault test, a scalar property caused `processFrontMatter` to throw, but the command still returned as executed and the error only showed up in dev tools.

One related thing to keep in mind: template/capture choices can also run inside macros. If frontmatter insertion errors are handled only locally with a notice, macro execution may still continue after a failed link insertion. I tested a small macro in the dev vault where step 1 hit the scalar-property frontmatter error and step 2 created a marker note; step 2 still ran. I don’t think this needs to be treated as a separate blocker by itself, but it’s a good reason to prefer awaiting/propagating the failure through the existing choice/macro error handling instead of making this a local-only notice.

The other bigger issue is YAML/frontmatter handling. At the moment this only works for missing/null properties and array properties. For example:

```yaml
related:
  - "[[Existing]]"
```

works, but:

```yaml
related: "[[Existing]]"
```

fails with `Could not add into non array property:related`.

That makes the feature feel more like “append to a frontmatter list property” than “insert into a frontmatter property”. I think we should either handle scalar values by converting them to a list, or make the UI/docs explicitly say that this only supports list properties. My preference would be a small helper with tests, roughly covering:

- missing/null property -> create `[newLink]`
- array property -> append `newLink`
- string/scalar property -> convert to `[existingValue, newLink]`
- object/map property -> reject with a clear error

I also found one settings issue: changing “Enabled (requires active file)” to “Enabled (skip if no active file)” drops `frontmatterProperty` from the saved `appendLink` object while keeping `placement: "inFrontmatter"`. After that, running the choice fails with `Invalid frontmatter property`. So the settings handlers need to preserve `frontmatterProperty` when rebuilding `appendLink`.

Suggested next steps:

1. Make the frontmatter insertion path async/awaited.
2. Extract the YAML value handling into a small tested helper.
3. Preserve `frontmatterProperty` across the append-link setting changes.
4. Add regression tests for the array, scalar, missing/null, and settings-preservation cases.
5. Clean up the small whitespace issue reported by `git diff --check`.

I’m happy to review another pass. The core idea is good; I just want this to be reliable across normal Obsidian frontmatter shapes before shipping it.

chhoumann:
Arjan, I owe you a correction here. I missed the June 9 and June 15 commits you added after my review comment, and those commits did address the important pieces I called out: async propagation, preserving the frontmatter settings, and handling the normal YAML value shapes.

I have now ported your broader settings-based feature into #1394 on top of current master and the shared append-link settings UI. The PR body credits your work directly because the explicit "In frontmatter property" placement and the configurable frontmatter handling are your feature direction, and it is the more complete product shape.

I am keeping this PR closed only because it now conflicts with master and I do not want to push to your branch. This is not a rejection of the work. You did a great job here, and thank you for continuing after the first review.

ArjanSeijs:
No worries, saves me some time from dealing with the merge conflicts :sweat_smile: .
Thank you for your work on this plugin!

chhoumann:
That’s very kind of you, thank you.

I've also added co-author attribution on the relevant commits so the Git history reflects your contribution.

I appreciate you taking the time to contribute to QuickAdd!

PR #1305, opened 2026-06-12
fix(package-import): classify bundled scripts as critical

chhoumann:
Classifies bundled script files as critical package-import capabilities even when the package's own command graph does not reference them.

This closes the capability-gate bypass where a package could drop or overwrite a `.js` script path and avoid the acknowledgement/review gate because `executable` only meant "referenced by this package." The `executable` field still keeps that command-graph meaning; the new `bundled-script` capability covers script-kind assets and `.js` destination paths as a separate review-required axis.

Verification:
- `bun run test -- packagePreview`
- `bun run test`
- `bun run lint`
- `bun run check`
- Dev vault proof: temporarily wired `dev/.obsidian/plugins/quickadd/main.js` to this worktree's generated bundle, reloaded QuickAdd, and ran `obsidian vault=dev quickadd:package-preview path=quickadd-plan006-proof.quickadd.json decode=true`. The live preview returned `critical=2`, two `bundled-script` critical rows, and `criticalScriptPaths` containing both the orphan `user-script` and the `.js` asset declared as `template`. Restored the original symlink to `/Users/christian/Developer/quickadd/main.js`, reloaded QuickAdd back, removed the temp package, and confirmed `data.json` hash was unchanged.

Notes:
- Root `main.js`/`styles.css` were generated only for dev-vault proof and are intentionally not part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Enhanced bundled script detection to flag files requiring review based on file type and installation path.
  * Improved package safety assessment and acknowledgement workflow for critical bundled files.

* **Tests**
  * Expanded coverage for script review behavior and package safety requirements.


chhoumann:
**Review (sanity-check pass): holding merge — Codex's P1 is confirmed and it's a release blocker for this feature.**

I traced the full chain at head `cececac`:

1. `buildPackagePreview` now puts orphan bundled scripts (declared script kinds and `.js`-path assets) into `criticalScriptPaths` even when `file.executable === false` (src/services/packagePreview.ts:818-822).
2. `isFullyReviewed` requires **every** path in `criticalScriptPaths` to be in `reviewedScriptPaths` (packagePreview.ts:951-958).
3. The only writer of `reviewedScriptPaths` is `FilePreviewRow.toggle()`, which fires `onReviewed` **only when `file.executable` is true** (src/gui/PackageManager/FilePreviewRow.svelte:54).

So any package with an orphan `user-script`/`conditional-script` or a `.js` asset declared as e.g. `template` can never satisfy the disclosure gate — expanding the row does nothing, and the Import button stays disabled forever. The capability-gate bypass is closed by bricking legitimate review instead of enabling it.

**Root cause is in Plan 006 itself**, which asserts "ImportPackageModal.svelte — No change needed — it already gates correctly off these values; this plan only widens what those values flag." That missed the `file.executable` condition inside `FilePreviewRow.toggle()`. The PR implemented the plan faithfully; the plan's no-UI-change claim was wrong.

**Suggested fix:** make the review predicate single-sourced in the pure module rather than duplicated in the UI — e.g. add a `requiresReview: boolean` to `PreviewFile` (computed in `buildPackagePreview` as `bundled && (executable || isExecutableBundledAsset(file))`, i.e. exactly the `criticalScriptPaths` membership predicate), and change FilePreviewRow to `if (expanded && file.requiresReview) onReviewed(file.originalPath)`. Then add a test asserting the invariant that every `criticalScriptPaths` entry corresponds to a file with `requiresReview === true`, so the gate can always be satisfied from the UI.

The preview-side classification itself looks right (kind-spoofing via `.js` path covered, `executable` keeps its command-graph meaning, severity/copy consistent with the other critical flags), and scope is otherwise clean. CodeRabbit is also still pending. Once the gate wiring is fixed + tested, this is good to merge.

chhoumann:
Re-reviewed at head `2684beef`: the gate-bricking P1 is fully resolved and the fix takes the single-source approach — `requiresReview` is computed once in `buildPackagePreview`, `criticalScriptPaths` is derived from exactly that predicate, and `FilePreviewRow.toggle()`/badge/skip-warning all key off `file.requiresReview`. The gate is satisfiable by construction now, and the new invariant test ("keeps critical script paths aligned with files that require review") plus the FilePreviewRow component test lock it in. Also verified the dropped `file.bundled` condition in the gate-set filter is vacuous: the files manifest is built only from `pkg.assets` with `bundled: true`, so every gate path has a renderable row. Will merge once CodeRabbit finishes on this head.

chhoumann:
@coderabbitai review

PR #1306, opened 2026-06-12
perf(ai): lazy-load js-tiktoken rank data

chhoumann:
## Summary

Lazy-load `js-tiktoken` rank data so QuickAdd no longer initializes the multi-MB `cl100k_base` / `o200k_base` tables on plugin startup.

Exact token counts remain on async request-gating paths. Synchronous UI/API surfaces now use a cheap `chars / 4` estimate so rendering token labels or calling `quickAddApi.ai.countTokens` does not force rank loading.

## Changes

- Replace eager tokenizer rank imports with dynamic rank loading and cached `Tiktoken` instances.
- Add `getTokenCountAsync` for exact counts and `estimateTokenCount` for synchronous preview paths.
- Keep OpenAI request budget checks and chunk sizing on exact async counts.
- Switch AI Assistant settings modals and public sync `countTokens` to the estimate path.
- Add focused token-count regression coverage and update the OpenAI request mock for the async count symbol.
- Update the current QuickAdd API docs so `countTokens` is documented as a fast estimate rather than an exact budget gate.
- Cache in-flight tokenizer construction promises so concurrent first callers share one lazy `Tiktoken` instance per encoding.

## Testing / validation

- `bun run test -- src/ai/OpenAIRequest`
- `bun run test -- src/ai/AIAssistant.tokenCount`
- `bun run lint`
- `bun run check`
- `bun run test`
- Structural checks:
  - no static `import ... from "js-tiktoken/ranks/*"` remains
  - two dynamic rank imports remain in `src/ai/AIAssistant.ts`
  - production bare `getTokenCount` usage is limited to `src/ai/AIAssistant.ts` and `src/quickAddApi.ts`
- Temporary production proof bundle built to `/private/tmp/quickadd-plan007-proof` without writing root `main.js` / `styles.css`; proof `main.js` was 4,423,058 bytes and contains lazy `Promise.resolve().then(...init...)` rank wrappers.
- Dev-vault proof with `vault=dev`:
  - before: `main.js -> /Users/christian/Developer/quickadd/main.js`
  - temporarily repointed `main.js` to `/private/tmp/quickadd-plan007-proof/main.js`
  - reloaded QuickAdd successfully
  - `plugin.api.ai.countTokens("abcde", { name: "gpt-4", maxTokens: 8192 })` returned `2`, matching the sync estimate
  - `quickadd:testQuickAdd` rendered the AI Assistant modal with `Token count: 1` and range max `8191`
  - `dev:errors` reported `No errors captured.`
  - after: restored `main.js -> /Users/christian/Developer/quickadd/main.js` and reloaded QuickAdd again
  - `data.json` SHA-256 stayed `3ffb1b1a9f25ddfdfa398d7b9c4a9d14e028069f8308b1d5806aa4845e7a3f86`

## Checklist

- [x] Linked any related issue(s). Plan 007 orchestration task; no GitHub issue.
- [x] Noted release/migration impact, if any. Perf-only runtime change; no migration. `quickAddApi.ai.countTokens` remains synchronous but is now a best-effort estimate.

## Summary by CodeRabbit

* **New Features**
  * Added async token counting API for precise token count calculations

* **Documentation**
  * Clarified that UI token estimates are rough approximations (~4 characters per token)
  * Updated guidance to display preflight warnings advising users to shorten text if estimates exceed limits

### DIFF
--- a/docs/docs/QuickAddAPI.md
+++ b/docs/docs/QuickAddAPI.md
@@ -585,17 +585,20 @@ const selectedModel = await quickAddApi.suggester(models, models);
 Gets the maximum token limit for a model.
 
 ### `countTokens(text: string, model: string): number`
-Counts tokens in text according to model's tokenization.
+Returns a fast synchronous token estimate (`~4` characters per token). This does
+not load tokenizer rank data and should be used for UI hints or rough preflight
+warnings, not as an exact model-limit gate. QuickAdd's built-in AI requests still
+use exact token counts internally before sending requests.
 
 **Example:**
 ```javascript
 const text = await quickAddApi.utility.getClipboard();
-const tokenCount = quickAddApi.ai.countTokens(text, "gpt-4");
+const estimatedTokens = quickAddApi.ai.countTokens(text, "gpt-4");
 
-if (tokenCount > 4000) {
+if (estimatedTokens > 4000) {
     await quickAddApi.infoDialog(
-        "Text Too Long",
-        `The text contains ${tokenCount} tokens, which exceeds the model limit.`
+        "Text May Be Too Long",
+        `The text is roughly ${estimatedTokens} tokens. Consider shortening it before sending.`
     );
 }
 ```
--- a/src/ai/AIAssistant.tokenCount.test.ts
+++ b/src/ai/AIAssistant.tokenCount.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+import type { Model } from "./Provider";
+import {
+	estimateTokenCount,
+	getTokenCount,
+	getTokenCountAsync,
+} from "./AIAssistant";
+
+const model = { name: "gpt-4", maxTokens: 8192 } as unknown as Model;
+
+describe("AI Assistant token counting", () => {
+	it("estimates tokens from character length synchronously", () => {
+		expect(estimateTokenCount("")).toBe(0);
+		expect(estimateTokenCount("abcd")).toBe(1);
+		expect(estimateTokenCount("abcde")).toBe(2);
+	});
+
+	it("keeps getTokenCount as the synchronous estimate wrapper", () => {
+		const text = "count this prompt quickly";
+
+		expect(getTokenCount(text, model)).toBe(estimateTokenCount(text));
+	});
+
+	it("counts tokens exactly through the lazy async encoder", async () => {
+		const first = await getTokenCountAsync("hello world", model);
+		const second = await getTokenCountAsync("hello world", model);
+
+		expect(Number.isInteger(first)).toBe(true);
+		expect(first).toBeGreaterThan(0);
+		expect(first).toBeLessThanOrEqual("hello world".length);
+		expect(second).toBe(first);
+	});
+});
--- a/src/ai/AIAssistant.tokenCountCache.test.ts
+++ b/src/ai/AIAssistant.tokenCountCache.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it, vi } from "vitest";
+import type { Model } from "./Provider";
+
+describe("AI Assistant token encoding cache", () => {
+	it("shares one tokenizer construction between concurrent first callers", async () => {
+		vi.resetModules();
+		const constructorMock = vi.fn();
+
+		vi.doMock("js-tiktoken/lite", () => ({
+			Tiktoken: class {
+				constructor(rank: unknown) {
+					constructorMock(rank);
+				}
+
+				encode(text: string) {
+					return Array.from(text);
+				}
+			},
+			getEncodingNameForModel: () => "cl100k_base",
+		}));
+		vi.doMock("js-tiktoken/ranks/cl100k_base", () => ({
+			default: { name: "cl100k_base" },
+		}));
+		vi.doMock("js-tiktoken/ranks/o200k_base", () => ({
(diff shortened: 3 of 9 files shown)


chhoumann:
**Review (sanity-check pass):** Implementation is faithful to Plan 007 and the lazy split is clean — request-gating paths (`OpenAIRequest` budget check, chunk sizing at AIAssistant.ts:577-642) all stay on exact `getTokenCountAsync`, UI labels and the sync public API go through the cheap estimate, and no static rank imports remain.

On the two AI-review themes:

1. **Public `countTokens` becoming an estimate (Codex P2 / Devin 🚩):** this is the deliberate Plan 007 decision ("keep the sync signature, make it honest as an estimate"), and the `8350c50` docs commit takes exactly the resolution path Codex offered (update the contract + example so scripts don't budget-gate on it). I consider this resolved — the alternative (async public API or eager ranks) defeats the plan's purpose. Worth a release-notes mention since existing scripts silently get estimates.

2. **Devin's `getEncodingAsync` race — confirmed, please fix before merge.** Concurrent first callers (e.g. RateLimiter dispatching several requests in ChunkedPrompt) each pass the empty-cache check, each `await loadRank()`, and each construct a multi-MB `Tiktoken`. No correctness impact (last writer wins, identical counts), but it momentarily multiplies exactly the allocation this PR exists to avoid. The standard fix is to cache the promise, not the instance:

```ts
const encodingCache = new Map<string, Promise<Tiktoken>>();

function getEncodingAsync(name: string): Promise<Tiktoken> {
	const encodingName =
		name === "o200k_base" || name === "cl100k_base" ? name : "cl100k_base";
	let cached = encodingCache.get(encodingName);
	if (!cached) {
		cached = loadRank(encodingName).then((rank) => new Tiktoken(rank));
		encodingCache.set(encodingName, cached);
	}
	return cached;
}
```

Holding merge for that fix + CodeRabbit/CI completing on the head commit. Everything else looks good to go.

chhoumann:
**Re-review at head `028ed8c0`:** the `getEncodingAsync` race fix is solid — promise-cached with error-path eviction so a failed load doesn't poison the cache, plus a concurrency regression test asserting a single `Tiktoken` construction. All code-level findings on this PR are now resolved, and CI is green (CodeRabbit re-processing after the push).

**However, I'm holding the merge on a scope/premise concern that deserves an explicit decision:**

1. **The plan's headline benefit is overstated.** On master, `new Tiktoken(rank)` — the expensive part that builds the ~100k-entry BPE Map — was *already* lazy behind `encodingCache`, constructed on first `getTokenCount` call. The multi-MB tables were never built at plugin startup. What this PR actually defers is the rank *module evaluation*: full-parsing ~3.2MB of bundled source (1.0MB cl100k + 2.2MB o200k) and materializing the raw rank strings.
2. **Bundle size is unchanged.** Obsidian requires a single-file `main.js` (esbuild cjs, no splitting), so the dynamic imports become inline lazy closures — the proof bundle is still ~4.4MB with both rank tables inside. No download/disk/file-read win.
3. **Real benefit:** deferring ~3.2MB of parse/eval out of plugin load (tens of ms desktop, more on mobile, where Obsidian's per-plugin load-time metric makes it visible) and a few MB of heap for users who never touch AI features. Real, but modest — and the PR pays for it with a public-API behavior change (`countTokens` exact → estimate), dual sync/async paths, and docs churn.
4. **The exactness being preserved is mostly illusory anyway:** every non-OpenAI model (Claude, Gemini, local) already falls back to `cl100k_base`, so the "exact" request gate is only exact for OpenAI tokenizers.

**Recommendation:** if startup cost is worth acting on, the bigger and simpler win is removing `js-tiktoken` entirely — estimate everywhere with a safety margin, let the provider API arbitrate overflow. That cuts the actual 3.2MB from the bundle (parse + memory + load win for every user) and deletes the complexity this PR adds. If instead OpenAI-exact gating is worth keeping, this PR is mergeable as-is — but with expectations downscoped to "defers ~3MB of parse at load," not "stops building multi-MB tables at startup."

Author's call — happy to merge on confirmation, or close in favor of a remove-tiktoken plan.

chhoumann:
Closing per the scope discussion above. Summary of the rationale:

- The expensive part (building the ~100k-entry BPE Maps via `new Tiktoken(rank)`) was **already lazy** on master behind `encodingCache` — the plan's premise that multi-MB tables were initialized at plugin startup was incorrect. This PR only defers evaluation of the rank data modules (~3.2MB of source parse + raw string materialization).
- Obsidian requires a single-file `main.js`, so the rank data ships in the bundle either way — no download, disk, or file-read win. The deferred-parse win is real but modest (tens of ms at plugin load, a few MB of heap for non-AI users).
- The "exact" token counting the PR works to preserve is only exact for OpenAI tokenizers; all other providers (Claude, Gemini, local models) already fall back to `cl100k_base` approximations.
- In exchange, the PR changes the public `countTokens` API from exact to estimate and adds dual sync/async code paths — complexity disproportionate to the benefit.

The better follow-up, if startup cost is worth acting on, is removing `js-tiktoken` entirely: estimate everywhere with a safety margin and let the provider API arbitrate overflow. That removes the actual 3.2MB from the bundle and simplifies the code rather than complicating it. The race-fix pattern and the concurrency test from this branch are worth carrying into that future PR.

(Plan 007 in `plans/` should be updated or retired to reflect this analysis.)

PR #1309, opened 2026-06-12
fix(file-suggester): register one document scroll listener per instance

chhoumann:
Refactor FileSuggester tooltip cleanup so each instance owns one document-level scroll listener instead of registering one listener per rendered suggestion row. Repeated suggestion batches still replace rows normally, but the scroll listener now lives on the suggester lifecycle and is removed by `close()`.

This fixes unbounded listener accumulation when typing through file-link autocomplete result batches, while preserving the 200ms hover tooltip behavior and owner-document safety for pop-out windows.

Verification:
- `bun run test -- src/gui/suggesters/fileSuggester.test.ts`
- `bun run test`
- `bun run lint`
- `bun run check`
- Structural greps: exactly one `addEventListener('scroll'` in `fileSuggester.ts`, no `_tooltipCleanup`
- Dev vault proof in `vault=dev`: temporarily pointed `main.js` at this worktree bundle, reloaded QuickAdd, drove `plugin.api.inputPrompt` with four 50-row `[[qa...` FileSuggester batches, observed one document scroll listener throughout, verified hover tooltip appears and document scroll removes it, then blur removed the single listener. Restored `/Users/christian/Developer/dev_vault/dev/.obsidian/plugins/quickadd/main.js` to `/Users/christian/Developer/quickadd/main.js` and reloaded QuickAdd afterward.

Generated artifacts:
- Ran `bun run build` only for dev-vault proof. It generated local `main.js` and `styles.css` in the worktree; they are ignored/uncommitted and were not edited by hand.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced file suggester tooltip behavior for improved reliability and proper event cleanup.

* **Tests**
  * Added test coverage to verify correct tooltip lifecycle management and event handling.

### DIFF
--- a/src/gui/suggesters/fileSuggester.test.ts
+++ b/src/gui/suggesters/fileSuggester.test.ts
@@ -680,6 +680,103 @@ describe('FileSuggester DOM XSS safety', () => {
         }
     });
 
+    it('registers exactly one document scroll listener regardless of row count', () => {
+        vi.useFakeTimers();
+        const frame = document.createElement('iframe');
+        document.body.appendChild(frame);
+        const popoutDocument = frame.contentDocument!;
+        const popoutWindow = frame.contentWindow!;
+        const inputEl = popoutDocument.createElement('input');
+        popoutDocument.body.appendChild(inputEl);
+
+        const runtimeFiles = new Map<string, TFile>();
+        const app = {
+            dom: { appContainerEl: document.body },
+            keymap: {
+                pushScope: vi.fn(),
+                popScope: vi.fn(),
+            },
+            workspace: {
+                getActiveFile: vi.fn(() => null),
+                on: vi.fn(() => ({})),
+            },
+            fileManager: {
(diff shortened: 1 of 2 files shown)


chhoumann:
**Review note on Devin's finding:** verified non-blocking. The document-level scroll listener this PR adds follows the base class's established lifecycle idiom — `TextInputSuggest` already registers document/window globals (pointerdown, wheel, resize, blur) in `open()` and removes them in `close()`, and `close()` fires through redundant paths (Escape via keyboard scope, input blur, outside pointerdown in capture phase, window blur/resize), which covers the Chromium edge where removing a focused element doesn't dispatch blur. The un-removed input/focus/blur listeners Devin notes are element-scoped (GC'd with the input element), pre-existing, and out of Plan 008's scope — Devin's own comment concedes both points. Holding merge only for CodeRabbit/Codex to finish.

PR #1314, opened 2026-06-12
chore(deps): move bundled runtime deps out of devDependencies

chhoumann:
Move five packages that are value-imported from src and bundled into QuickAdd from devDependencies to dependencies: @popperjs/core, obsidian-dataview, svelte-dnd-action, three-way-merge, and uuid. The workspace declaration in bun.lock is updated to match package.json.

This is metadata-only: dependency versions are unchanged, resolved lockfile package entries are unchanged, and generated plugin artifacts such as main.js and styles.css were not built or modified.

Verification:
- Drift check: package.json and bun.lock had no baseline drift; CaptureChoiceEngine.ts had unrelated in-scope drift, but the three-way-merge value import still matched the plan evidence.
- grep -rn '@popperjs/core\|obsidian-dataview\|svelte-dnd-action\|three-way-merge\|"uuid"' src/
- node -e dependency-scope check printed OK
- bun install --frozen-lockfile
- bunx tsc -noEmit -skipLibCheck
- bun run lint
- bun run test
- bun run check

## Summary by CodeRabbit

* **Chores**
  * Updated dependency configuration to move several packages from development dependencies to runtime dependencies, ensuring required libraries are properly classified for production use.


chhoumann:
**Review note on Devin's two analysis flags — both anticipated by Plan 016's determination rule, non-blocking:**

1. *Community convention (all-devDeps):* the plan's motivation is precisely that the convention misleads audit tooling — `npm audit --omit=dev`, Renovate grouping, and this repo's own Dependency Review CI gate all treat devDependencies as non-shipping, so the five bundled packages were escaping production-grade scrutiny. Since these are value-imported from `src/` and shipped inside `main.js`, classifying them as `dependencies` reflects reality; the build output is unaffected either way.
2. *`svelte` staying in devDependencies:* deliberate per the plan's rule — the boundary is "value-imported from `src/`". `svelte` is consumed as a compiler; its runtime enters the bundle via compiled `.svelte` output, not via a source import, and a production-only install never happens for a bundled plugin, so the peer-dependency concern is moot.

CI (including the Dependency Review gate this change feeds), CodeRabbit, and the frozen-lockfile install are all green. Merging.

Issue #964, opened 2025-10-14
Member Access Picks First User Script Without Warning


chhoumann:
## Overview
Triggering `MacroName::export` currently short-circuits to the first user-script command in the macro. Commands before that script run, its export executes, and any later user scripts are ignored. This makes macros with multiple scripts behave unpredictably and gives no feedback to the user about what happened.

## Why It Hurts
- Macro authors expect the export matching the suffix they typed, not whatever script appears first.
- Multi-script macros (variants, pipelines) can’t expose the right export reliably; the wrong code runs silently.
- There’s no UI, doc, or log hint describing the limitation, so it looks like the feature is broken.

## Reproduction
1. Create a macro with two user-script commands.
2. Give the second script an export named `beta`.
3. Run `MacroName::beta`.
4. Observe that only the first script becomes eligible for the export.

## Desired Experience
- Make it explicit which user script provides member-access exports.
- Leave single-script macros frictionless.
- Surface a clear notice when a macro isn’t configured for member access instead of guessing.

## Possible Approaches
1. Let authors mark exactly one user-script command as "Expose for member access" in the macro editor; auto-select when only one script exists.
2. At runtime, if no script is marked (or multiple are), show a notice guiding the user to configure the macro instead of picking one.
3. Update tooltips/docs to spell out the rule (`Macro::export` runs the designated user script export and ignores others).
4. Optionally log a warning when multiple scripts exist but none is designated, helping users catch misconfiguration in tests.

## Open Questions
- Should designation live on the macro or the individual command?
- How do we migrate existing macros gracefully (auto-designate the sole script?)

chhoumann:
Closing — the reported behaviour was **fixed and shipped in 2.12.1** and verified in a live vault on the current release (2.13.x).

`{{MACRO:Name::member}}` no longer picks the first user script. It searches all user-script commands in the macro and:
- runs the unique exporter of the requested member;
- aborts with a clear message (not a silent guess) when **no** script exports it;
- aborts and lists the conflicting script names when **multiple** scripts export it, telling you to disambiguate with `{{MACRO:Name::<Script Name>::member}}`.

Reproduced live (2.13.1):
| Call | Result |
|---|---|
| `Macro::beta` where only the **second** script exports `beta` | resolves to the second script ✅ |
| `Macro::beta` where two scripts export `beta` | clear "multiple user scripts exporting 'beta'" error + disambiguation hint ✅ |
[...]

PR #1326, opened 2026-06-13
fix(macro): harden user-script member access conflicts (#964)

chhoumann:
## Summary

Follow-up **hardening** for macro member access (`{{MACRO:Name::member}}`). The original #964 bug — member access silently short-circuiting to the *first* user script — was already fixed and shipped in **2.12.1** (merged-namespace model in `SingleMacroEngine.ts`). This PR fixes three residual issues that investigation + adversarial review surfaced in that model, scoped to the safe, regression-free subset.

### What changes

- **F5 — convention-key conflict (the one real regression).** Since 2.12.1, `{{MACRO:M::settings}}` *hard-aborts* whenever two or more user scripts export the same member. `settings` is a documented convention export (consumed by `initializeUserScriptSettings`) and `entry` is the object-export entrypoint — both are exported by many scripts, so this broke macros that worked before 2.12.1. Now the **reserved convention keys `settings` and `entry`** resolve to the **first-declared exporter** (pre-2.12.1 behavior) and surface a **one-time notice** pointing at the `::<Script Name>::` selector. Genuine entrypoint conflicts (`start`, `run`, custom names) still hard-abort — the deliberate “abort instead of guessing” behavior is preserved.
- **F6 — candidate refresh by id.** After pre-commands run, the selected user-script command is re-resolved by `command.id` (not by numeric index), and the post-command slice is derived from the refreshed index. Hardens against a pre-command mutating the macro's command list (latent today; defensive).
- **F4 — no silent empty string.** Requesting a member on a macro that has no user scripts now logs a clear warning instead of silently resolving to `""`.

Deliberately **out of scope** (deferred — riskier than the bugs they fix, see review notes): F1/F2 (probe executes every script's module body — fixing via a load-cache risks *stale* execution), F3 (selector fall-through), F7 (unifying the `getUserScript` `command.name` grammar — that grammar is load-bearing for the documented `myScript::start` command form).

## Reproduction / verification

Verified end-to-end in a real Obsidian dev vault (v2.13.1) via `api.format(...)`:

| Scenario | Result |
|---|---|
| `::settings` with 2 exporters | resolves to first exporter **+ notice** (was: hard abort) |
| `::beta` ambiguous (non-reserved) | still aborts (deliberate behavior preserved) |
| single-script `::beta` | unchanged |
| later-script `::beta` | `SECOND_BETA` (original #964 fix intact) |
| `::start` on a script-less macro | empty **+ warning** (was: silent `""`) |

`pnpm test` (1816 pass), `pnpm lint`, and `tsc --noEmit` all green; added 4 unit tests (reserved-key resolution, preserved non-reserved abort, zero-user-script warning, id-based refresh under command-list mutation).

## Release / migration impact

`fix:` → patch release. No data migration. Behavior change is limited to `{{MACRO:M::settings}}` / `{{MACRO:M::entry}}` when 2+ scripts export them (now resolves + warns instead of aborting) and a new diagnostic notice for script-less member access.

Refs #964 (literal bug fixed in 2.12.1; this is the residual hardening).

## Summary by CodeRabbit

* **Documentation**
  * Expanded macro member-selection docs with warnings and reserved-key behavior.

* **Bug Fixes**
  * Reserved convention keys (`settings`, `entry`, `quickadd`) now pick the first matching script and emit a one-time notice pointing to the selector option.
  * Ambiguity for non-reserved members still errors as before.
  * Macros that export no user-script commands now warn and return an empty result.
  * Script selection remains correct when pre-commands modify the command list.

### DIFF
--- a/docs/docs/Choices/MacroChoice.md
+++ b/docs/docs/Choices/MacroChoice.md
@@ -412,12 +412,15 @@ that uniquely exports the requested member across all scripts in the macro.
 QuickAdd resolves `Macro::member` like this:
 - If exactly one script exports the requested member, QuickAdd uses it.
 - If no script exports the requested member, QuickAdd stops and shows an error.
+  - Exception: if the macro has no user-script commands at all, QuickAdd cannot
+    satisfy member access — it logs a warning and returns an empty result instead
+    of stopping the macro.
 - If multiple scripts export the requested member, QuickAdd stops and lists the
   conflicting script names instead of guessing.
-- Exception: the convention keys `settings` and `entry` (which many scripts export)
-  resolve to the **first** script that exports them and show a one-time notice
-  pointing at the selector form below, rather than stopping. Use the selector if you
-  need a different script.
+- Exception: the convention keys `settings`, `entry`, and `quickadd` (which many
+  scripts export as metadata rather than entrypoints) resolve to the **first** script
+  that exports them and show a one-time notice pointing at the selector form below,
+  rather than stopping. Use the selector if you need a different script.
 
 If there is a conflict, you can target a specific script by name:
 - `{{MACRO:MyMacro::Script 1::option1}}`
--- a/docs/docs/Choices/MacroChoice.md
+++ b/docs/docs/Choices/MacroChoice.md
@@ -414,6 +414,10 @@ QuickAdd resolves `Macro::member` like this:
 - If no script exports the requested member, QuickAdd stops and shows an error.
 - If multiple scripts export the requested member, QuickAdd stops and lists the
   conflicting script names instead of guessing.
+- Exception: the convention keys `settings` and `entry` (which many scripts export)
+  resolve to the **first** script that exports them and show a one-time notice
+  pointing at the selector form below, rather than stopping. Use the selector if you
+  need a different script.
 
 If there is a conflict, you can target a specific script by name:
 - `{{MACRO:MyMacro::Script 1::option1}}`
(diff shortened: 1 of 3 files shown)


chhoumann:
### Adversarial review pass (Codex skeptic + architect on the final diff)

No high-severity blockers. Confirmed: F6 runtime aborts correctly when the selected command's id disappears mid-run; F5 first-exporter ordering is correct; the conflict notice flag resets per run; and the diff does **not** touch the deferred paths (`getUserScript`, `MacroChoiceEngine`, `resolveScriptSelector`) so F1/F2/F3/F7 stay out of scope.

**Addressed in `c952a24`:**
- **Reserve `quickadd` too** (Medium). It's a third QuickAdd-consumed convention key (`collectChoiceRequirements` reads `userScript.quickadd.inputs`), so `::quickadd` now resolves to the first exporter like `settings`/`entry` instead of hard-aborting.
- **F4 false-positive** (Low). Member resolution now uses the `{found, value}` distinction, so a member that exists but holds `undefined` no longer warns as "missing." Folded the duplicate `applyMemberAccess` into `resolveMemberAccess`.
- **F6 test strengthened** (Low). The "command removed mid-run" test now has a trailing user script, so a buggy stale-index fallback would route to that neighbour — the test actually distinguishes the id-based refresh now.

**Consciously not changed:**
- Arbitrary command **reorder** by a pre-command (Medium) — F6 handles insert-before / removal (the realistic mutations); full reorder safety is out of scope for this latent hardening and no current command mutates the running macro's command list.
- The conflict notice is one-time **per macro-token run**, not per template (a fresh engine is spawned per `{{MACRO:…}}`); repeated identical tokens can re-notice. Accepted.

`pnpm test` (1818 pass), `pnpm lint`, `tsc --noEmit` all green.

PR #1330, opened 2026-06-13
fix(gui): preserve choice-builder edit position by converting the ChoiceBuilder trio to Svelte (#1130)

### LINKED ISSUE #1130: [FEATURE REQUEST] UX Improvement: Modal completely refreshes on button clicks, causing loss of edit position
I'm always frustrated when I click certain buttons inside the QuickAdd modal (such as in the Macro Manager or settings). The entire modal completely refreshes and re-renders, which resets the scroll position and causes me to lose track of exactly where I was editing. 

I would like the related UI rendering code to be optimized so that clicking these buttons updates the modal without a full refresh. If a complete re-render is absolutely necessary, it would be great if the plugin could automatically save and restore the previous scroll position and focus state so the view doesn't jump.

Currently, the only workaround is to manually scroll back down and visually search for the item I was just editing after every single click, which is highly inefficient and interrupts the workflow.

**Additional context**
As shown in the attached GIF, you can see how the entire modal jumps and refreshes when the buttons are clicked.  If you are interested in addressing this, I would be happy to help by identifying and listing out all the specific buttons that trigger this refresh.

Image

chhoumann:
Fixes #1130.

## Problem

The Capture/Template choice-builder modals fully refreshed (`contentEl.empty(); display()`) on every toggle/dropdown change, resetting scroll position and dropping input focus/caret. Reported still-broken in 2.12.0.

A prior attempt (#1132–#1136, an `xstate` reload machine that snapshotted/restored scroll+focus) was reverted in #1137 — we didn't want that architecture. The Svelte 5 migration (#1248) shipped the `mountComponent` seam but left these imperative `new Setting()` builders as backlog.

## Approach

Convert the **ChoiceBuilder trio** (`choiceBuilder.ts` base + `captureChoiceBuilder.ts` + `templateChoiceBuilder.ts`) from imperative Obsidian `Setting` builders into Svelte 5 runes components mounted through the existing `mountComponent` seam. Conditional rows become reactive `{#if}` blocks, so **`reload()` is eliminated** — the modal never tears down, so scroll and caret are preserved by construction. No `xstate`, no scroll-restore band-aid, no new deps.

This is the first concrete slice of the imperative-GUI → Svelte migration; the shared primitives are reusable for the remaining modals.

## What changed (5 commits, each independently green)

1. **Shared primitives** — `SettingItem`/`Toggle`/`Dropdown` emit Obsidian-exact `.setting-item` / `.checkbox-container` / `.dropdown` markup (zero new CSS, native theming); `ValidatedInput` (port of `validatedInput.ts` incl. the `validateToken` staleness guard); `FormatPreviewField` (un-debounced, parity); `AppendLinkSetting` (collapses the duplicated capture/template logic), `FileOpeningSetting`, `OnePageOverrideSetting`, `OpenFileSetting`, `ChoiceNameHeader`; a `use:` suggester action.
2. **TemplateChoiceBuilder** → `TemplateChoiceForm.svelte`.
3. **CaptureChoiceBuilder** → `CaptureChoiceForm.svelte` + section components (`CaptureTargetSetting` + keyed `CanvasNodePicker`, `WritePositionSetting`, `InsertAfter/BeforeFields`); pure canvas helpers extracted verbatim to `canvasNodes.ts`.
4. **Base** reduced to a thin Svelte mount shell (removes `reload()` + all `add*` helpers + the dead `addFileSearchInputToSetting`).
5. **Review fixes** (see below).

## Persistence boundary (subtle, load-bearing)

A Svelte `$state` proxy does **not** write through to the original object (`proxy.js`), so `ChoiceBuilder.onClose()` resolves `snapshot(getResultChoice())`, and converted builders override `getResultChoice()` to return the form's proxy (`this.formProps.choice`) — not `this.choice`. Snapshotting the original would silently drop every edit. Callers already spread the result, so resolving a plain snapshot is safe.

## Adversarial review (Codex) found + fixed two real issues

- **HIGH** — newly *created* choices are class instances (`new CaptureChoice()` from `createChoice()`), and `$state()` does **not** deeply proxy class instances (only `Object`/`Array` prototypes). So in the add-new flow, nested toggles wouldn't reveal conditional rows (existing choices load as plain JSON, which is why it hid). Fix: `structuredClone` the choice in the form-props factories. Regression test uses a real `CaptureChoice` instance and **fails without the clone**.
- **MEDIUM** — `InsertBeforeFields` didn't default `createIfNotFound`/`createIfNotFoundLocation` (parity gap vs `InsertAfterFields`), risking a persisted `undefined` that throws in the capture formatter. Fixed by defaulting at init.

## Verification

- **Gates:** `tsc`, `eslint`, `svelte-check` 0/0, `vitest` **1835 pass** (incl. new component + persistence + class-instance regression tests), production esbuild build clean.
- **Dev-vault e2e (real Obsidian):**
  - Editing an existing Capture choice: scrolled to bottom (scrollTop 556), focused the format textarea with caret mid-string, changed the write-position dropdown (previously `reload()`'d) → **scrollTop stayed 556, focus + caret intact, the dependent section appeared reactively, no errors.**
  - Toggle → close → **edit persisted to `data.json`** (proves the proxy-snapshot path).
  - **Add-new flow:** New choice → Capture → fresh builder; toggling "Create file if it doesn't exist" and write-position "After line…" revealed their conditional rows reactively (confirms the class-instance fix).
  - Both builders mount with correct structure/theming and no console errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Pick specific canvas nodes as capture targets with capturability checks and helpful messages.
  * Redesigned reactive configuration forms with validated inputs, live format/filename previews, suggesters, and improved conditional sections.
  * New capture controls: write-position selector, insert-before/after fine‑tuning, append-link options, folder/file chooser, one-page override, and file-opening settings (location, split, view, focus).
* **Tests**
  * Added unit tests covering form reactivity, validation, and conditional rendering.


chhoumann:
Addressed review feedback in 841c7f67:

- **Codex P2 (file-opening normalize):** `FileOpeningSetting` now normalizes `fileOpening` at init, so an imported/legacy choice with `openFile:false` and no `fileOpening` object no longer throws when Open is toggled on (same point the imperative `addFileOpeningSetting` normalized). Regression test added. 👍
- **CodeRabbit (SettingItem):** renders `(control ?? children)` instead of both slots.
- **CodeRabbit (CanvasNodePicker):** shows an accurate "selected node was not found" message once nodes are loaded, instead of "still loading".

Gates green: tsc, eslint, svelte-check 0/0, vitest 1884.

PR #1335, opened 2026-06-13
feat(capture): order note picker by recency like Quick Switcher (#745, #539)

### LINKED ISSUE #539: [FEATURE REQUEST] Leverage Quick Switcher UX when selecting note to Append
I would love to follow the behaviour of Obsidian default quick switcher when selecting notes to append in QuickAdd. 

Current behaviour:
1. Top selected item isnto select a Note that doesn't exist. This is true even when you have disabled "Create note if it doesn't exist". 
2. Shows files from Scripts folder 
Screenshot_20230722_183651_Obsidian

Desired behaviour:
1. When searching for a note, default selection to most relevant note 
2. Once you type text that doesn't match any note, offer to create a new note.
Screenshot_20230722_183359_Obsidian
Screenshot_20230722_183351_Obsidian
Screenshot_20230722_183328_Obsidian

If possible, bonus features that come from Quick Switcher would be:
- Respect excluded files 
- Show files based on recency (either from QuickAdd Append, or from Obsidian recent files)

### LINKED ISSUE #745: [FEATURE REQUEST] Quick add find existing note prompt shows recent or frequently used files.
CleanShot 2024-10-10 at 13 08 46@2x

When using Quick Add to find an existing file, it seems to only show my "Scrpts" folder which I never access.

Ideal scenario would be to show recent or frequently used files.

Thank you!

chhoumann:
## What & why

Closes #745 · Closes #539

The capture file picker listed notes in **vault-registry order**, so it surfaced whatever sorted first (often a `Scripts` folder no one opens) instead of the notes you actually use — and typing a name offered a confusing "create" row even for notes that exist. This makes the picker mirror Obsidian's **Quick Switcher**:

- **Recently-opened notes first**, then a predictable **alphabetical** tail; **Excluded files** sink to the bottom (still selectable).
- Typing fuzzy-ranks the best match to the top (press Enter to pick it).
- A gated, clearly-labelled **"Create new note: X"** row that only appears when *Create file if it doesn't exist* is on **and** the typed name isn't an existing note.
- The **tag** picker now shows note names instead of raw `Some/Deep/Folder/Note.md` paths.

## How

- **New pure `orderFilesForPicker`** (`src/utils/fileOrdering.ts`): excluded sink → session recency (`FileIndex.openedAt`) → cross-session recents (`getLastOpenFiles`, cold-start seed) → alphabetical. Deliberately **not** `mtime` (sync/import/Templater bumps it and would re-surface noise). All Obsidian access is injected (`buildPickerOrderingDeps`), so the logic is unit-testable and degrades to alphabetical on older/missing APIs.
- **Recency reuse, no startup cost**: `FileIndex.getInstanceIfExists()` (never constructs) + `getLastOpenedAt()` via a new non-mutating `LRUCache.peek()`.
- **`InputSuggester`** gains **opt-in** `allowCustomValue` / `customValueLabel` / `valueExists`; defaults preserve today's behaviour, so the 6 non-capture callers are untouched. Create-row suppression now checks `displayItems` + `valueExists`, fixing a case where an existing in-folder note typed by basename showed a mislabelled "create" row.
- Wired into the **folder, vault, and tag** pickers in `CaptureChoiceEngine` plus the **one-page preflight** dropdown.

## Verification

- `tsc` + `eslint` clean; **full vitest 1901 passed** (new: `fileOrdering`, InputSuggester gating/suppression, engine ordering+gating, `FileIndex` recency seam; import-cycle invariant green).
- **Live dev-vault e2e**: the picker's empty state now renders recents-first (`QuickAdd-Repros/662-native.md`, `QA-620-GVAR-Output.md`, `Templates/Games Template.md`, …) instead of the old `qa-trace-*` registry junk — verified equal to the expected recency order, not the raw `getMarkdownFiles()` order.
- Multi-agent review (design + post-implementation adversarial passes); all surfaced blockers/majors addressed.

## Notes / not in scope

- No new setting (no migration). This **changes the default order** — worth a release-note callout. Today's order is arbitrary registry order, so there's no alphabetical contract being broken.
- Pre-existing behaviours left as-is: tag-scope "create" makes a vault-root note that won't carry the tag; the empty-folder invariant aborts before the picker even when *create* is on.

## Summary by CodeRabbit

* **New Features**
  * Suggestions now order by recent opens first, then alphabetically.
  * Pickers gain options to enable/disable a custom "create" row and to customize its label.
  * Note-path suggestions render clearer title/parent metadata.

* **Bug Fixes**
  * "Create new note" is suppressed when an equivalent target already exists (folders and tags).
  * Picker ordering correctly handles recency tiers and excluded files.

* **Tests**
  * New unit tests cover create-row behavior, picker ordering, and file-recency accessors.

### DIFF
--- a/src/engine/CaptureChoiceEngine.selection.test.ts
+++ b/src/engine/CaptureChoiceEngine.selection.test.ts
@@ -413,6 +413,81 @@ describe("CaptureChoiceEngine capture target resolution", () => {
 		expect(resolved).toBe("Inbox/note.md");
 	});
 
+	it("orders the folder picker by recency and gates the create row when enabled", async () => {
+		vi.mocked(getMarkdownFilesInFolder).mockReturnValue([
+			{ path: "Inbox/Apple.md", basename: "Apple" },
+			{ path: "Inbox/Zebra.md", basename: "Zebra" },
+			{ path: "Inbox/Mango.md", basename: "Mango" },
+		] as any);
+
+		const suggestSpy = vi.fn(async () => "Inbox/Apple.md");
+		(InputSuggester as any).Suggest = suggestSpy;
+
+		const app = createApp() as any;
+		app.workspace.getLastOpenFiles = () => ["Inbox/Zebra.md"];
+
+		const engine = new CaptureChoiceEngine(
+			app,
+			{ settings: { useSelectionAsCaptureValue: false } } as any,
+			createChoice({
+				captureTo: "Inbox/",
+				createFileIfItDoesntExist: {
+					enabled: true,
+					createWithTemplate: false,
+					template: "",
+				},
+			}),
+			createExecutor(),
+		);
+
+		await (engine as any).selectFileInFolder("Inbox/", false);
+
+		expect(suggestSpy).toHaveBeenCalledTimes(1);
+		const [, displayItems, items, options] = suggestSpy.mock
+			.calls[0] as unknown as [
+			unknown,
+			string[],
+			string[],
+			{
+				allowCustomValue: boolean;
+				customValueLabel: (value: string) => string;
+			},
+		];
+
+		// Recently opened (Zebra) first, then the rest alphabetically.
+		expect(items).toEqual([
+			"Inbox/Zebra.md",
+			"Inbox/Apple.md",
+			"Inbox/Mango.md",
+		]);
+		expect(displayItems).toEqual(["Zebra.md", "Apple.md", "Mango.md"]);
+		expect(options.allowCustomValue).toBe(true);
+		expect(options.customValueLabel("New")).toBe("Create new note: New");
+	});
+
+	it("disables the create row when create-if-not-exists is off", async () => {
+		vi.mocked(getMarkdownFilesInFolder).mockReturnValue([
+			{ path: "Inbox/Apple.md", basename: "Apple" },
+		] as any);
+
+		const suggestSpy = vi.fn(async () => "Inbox/Apple.md");
+		(InputSuggester as any).Suggest = suggestSpy;
+
+		const engine = new CaptureChoiceEngine(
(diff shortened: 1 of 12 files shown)


chhoumann:
### Review follow-up (multi-agent adversarial + ultracode pass)

A deeper review round caught one **real regression** in the first commit, now fixed in `cae610d2`:

- The capture create-row suppression added a `displayItems.includes(customValue)` check that ran for the *shared* `InputSuggester`. For the public `api.suggester` and the documented `{{VALUE:items|text:labels|custom}}` format syntax (display labels ≠ values, `allowCustomInput` on), this silently dropped a typed value equal to a display label. Fixed by relying on `valueExists` (capture-only) instead, with a regression test locking the generic-caller contract.

Also folded in two cheap correctness nits: `captureTargetExists` now probes a `.canvas` sibling, and `buildPickerOrderingDeps` hardens against a non-array `getLastOpenFiles`.

#### Deliberately deferred (tracked, not regressions)
- **Tag-scope "Create new note"** creates a note at vault root that does **not** carry the scoping tag (so it won't reappear in that tag picker). Pre-existing behaviour that the new labelled affordance makes more visible; proper fix is tag injection on create. Follow-up.
- **Empty folder/tag aborts before the picker** even when *Create file if it doesn't exist* is on, so the create row is unreachable in exactly the empty case. Pre-existing invariant (predates this PR); making it create-aware is a separate enhancement.
- **One-page preflight dropdown** gets the recency ordering but no create affordance / basename rows — the runtime picker and one-page mode diverge for the same choice. Pre-existing divergence.
- **Folder/vault pickers** keep full-path rows with fuzzy-match highlight; only the tag picker gets the basename + folder two-line rows. Cosmetic.

Happy to spin any of these into their own issues.

PR #1337, opened 2026-06-13
feat(templates): support format syntax in the template path (#620)

### LINKED ISSUE #620: Allow format syntax in template path input [FEATURE REQUEST]
I'm automating a way for me to dynamically add a collection. 
> /Games <- Collection Name
> /Games/Games.md <- Root File
> Templates/Games Template.md <- Template File for its items
>> The one above but games as dynamic function

Using `Templates/{{value:collectionName}} Template.md` as Template Path is resulting in an error
`Template file not found at path "Templates/{{value:collectionName}} Template.md".`

chhoumann:
Closes #620

## What & why
Using QuickAdd format syntax in a Template choice's **Template Path** (e.g.
`Templates/{{value:collectionName}} Template.md`) previously failed with
`Template file not found` — the path was the one note-creation input never run through the
formatter (file name, folder, and body already were). This makes the template path dynamic,
extending the recently-shipped `{{FOLDER}}`/`{{VALUE}}` work.

## Approach
Resolve the path **once at each engine's entry** and thread the resolved value into **both**
target-path construction (extension/name) **and** the content read — so the file that's created
and the template that's read can never disagree (a token expanding to `.canvas` creates a canvas
file, not `.md`). `getTemplateContent` takes a pre-resolved path and no longer formats (avoids
double-evaluating `{{date}}`/`{{random}}`).

New `CompleteFormatter.formatTemplateFilePath` is a deliberately **path-safe** subset of `format()`:
resolves value/date/time/field/global/selected/clipboard/vdate/random/math, but does **not** run
macros, inline JavaScript, or `{{TEMPLATE:}}` inclusion (a file-path lookup shouldn't execute code).
It guards `{{title}}` (incl. when injected via a global var), leaves `{{FOLDER}}`/`{{FILENAMECURRENT}}`
literal, and trims.

## Covered flows
- Template choice — create / overwrite / append (resolved path drives the target extension/name).
- Capture choice — *create file with template*.
- Insert template into the active note + the move/rename offer.
- Apply-template-to-note & the API — now **re-validate the resolved extension** so a token that
  resolves to `.canvas`/`.base` is not applied to a markdown note.
- Preflight (one-page input modal + CLI `quickadd:check`/`quickadd:run`) — collects tokens in the
  template path via a shared `scanTemplateSource` (template choices and capture create-with-template).
- Choice builder — shows a neutral "Contains format syntax — resolved at run time" hint instead of a
  false "Template not found".

## Tests
`pnpm test` (1901 pass), `pnpm lint`, `pnpm build`, and `cd docs && pnpm build` all green. Added unit
coverage for `formatTemplateFilePath` (path-safe allowlist, `{{title}}` guard incl. global-injected,
trim), the resolved-extension match in `TemplateChoiceEngine`, preflight path-token collection
(template + capture), and the `hasTemplatePathSyntax` helper.

## Verified in the dev vault (e2e)
Real Obsidian, current build:
- `quickadd:check` reports `collectionName` (the path token) as a required input.
- `quickadd:run id=… value-collectionName=Games` creates the output file from the resolved
  `Templates/Games Template.md` (and the body's `{{value:collectionName}}` reuses the same answer).

## Docs
Documented on the Template choice page (both `next` and the served `version-2.13.1` snapshot),
including the path-safe restrictions and the known limitations below.

## Known limitations (documented, intentional)
- `{{TEMPLATE:}}` inclusion with a *tokenized* path isn't supported (the token regex can't capture a
  path containing `}}`).
- A dynamic template path can't be auto-bundled when exporting a QuickAdd package (it isn't a literal
  file).
- One-page preflight can't pre-scan the *body* of a template whose path is dynamic (the path depends
  on answers still being collected); those prompts surface at run time.
- Anonymous `{{value}}` (vs named `{{value:x}}`) in a path may re-prompt across the capture→template
  boundary.

## Review
Design and implementation were each adversarially reviewed (opposite-model + multi-lens). The seam
choice (resolve-at-entry vs. formatting inside the read) and the resolved-extension validation came
directly from that review.

## Summary by CodeRabbit

* **New Features**
  * Template paths support QuickAdd format syntax and are resolved once at run time so file extension and collision behavior match the resolved template.

* **Bug Fixes**
  * Applying a template re-checks the resolved path and prevents non-markdown templates, showing a notice instead of proceeding.

* **Validation**
  * Validators detect template-path syntax and display a neutral “resolved at run time” hint rather than performing existence checks.

* **Documentation**
  * Clarified supported tokens, path-safe formatting rules, and limits for dynamic template paths.

* **Tests**
  * Added coverage and mocks for template-path formatting, resolution, and preflight scanning.


chhoumann:
### Review summary

This PR was reviewed in three passes (multi-lens ultracode workflow + opposite-model adversarial reviewers on the design, the implementation, and the final diff). No merge-blockers were found; the verified findings were addressed in `5befe327`:

- **Neutral builder hint rendered error-red.** The new "Contains format syntax — resolved at run time" hint reused `.qa-field-hint` (always `--text-error`). Added a `.qa-field-hint--neutral` modifier so a valid+message result shows muted, not red. (+`ValidatedInput` test)
- **`{{title}}` re-check ordering.** Moved the post-global-expansion guard to run immediately after global expansion, before user-input substitution, so a global-injected `{{title}}` still errors clearly without false-positiving on a user value that contains literal `{{title}}`.
- **Preflight traversal bug.** `scanTemplateSource` now snapshots + clears the shared `templatesToScan` set before recursing; the previous walk mutated the set it was iterating, which could drop nested/sibling literal templates.
- **Docs versioning.** Dropped the unreleased-feature note from the stable `version-2.13.1` snapshot (it ships in the version that contains it); kept it in the unreleased docs and documented the note-relative tokens (`{{FOLDER}}`/`{{FILENAMECURRENT}}`/`{{LINKCURRENT}}`) left literal in a path.
- **Test gap.** Added a regression test for the apply-to-note resolved-extension guard (a markdown-looking path that resolves to `.canvas`/`.base` is rejected).

New behavior worth calling out: Capture *create-with-template* bodies are now pre-scanned at preflight via the shared `scanTemplateSource` (verified safe — variables are shared through the choice executor, so no re-prompting).

Verified: `pnpm test` (1903 pass), `pnpm lint`, `pnpm build`, and `cd docs && pnpm build` all green; dev-vault e2e confirms `quickadd:run id=… value-collectionName=Games` creates the file from the resolved path and `quickadd:check` collects the path token.

chhoumann:
@coderabbitai thanks — addressed the nitpicks:

1. **Note-relative tokens in docs** — already documented in `docs/docs/Choices/TemplateChoice.md` (commit 5befe327): `{{FOLDER}}`/`{{FILENAMECURRENT}}`/`{{LINKCURRENT}}` are noted as left-literal in a template path. I intentionally did **not** mirror this into `version-2.13.1` — that's the stable snapshot for an already-released version that does not contain this feature, so documenting it there would mislead 2.13.1 users. It'll ride into the next version's snapshot at release.
2. **`resolvedTemplatePath` param naming** — done (this commit): renamed in `createFileWithTemplate`/`overwriteFileWithTemplate`/`appendToFileWithTemplate`.
3. **kebab-case filename** — skipping: the closest sibling `src/utils/valueSyntax.ts` (same `*Syntax.ts` pattern, and many other `src/utils/*.ts` like `fileOrdering.ts`, `pathValidation.ts`) use camelCase, so `templatePathSyntax.ts` matches the de-facto local convention; renaming it alone would make it inconsistent with its neighbors.

PR #1371, opened 2026-06-16
feat(format): date snap options |startof:/|endof: for {{DATE}} and {{VDATE}} (#511)

### LINKED ISSUE #511: [FEATURE REQUEST] Option for Week-Month in {{DATE:}}
I use a planner following the naming scheme `gggg.MM.[Wk]w.md`. So:
```
2023.05.Wk19.md
2023.05.Wk20.md
2023.05.Wk21.md
2023.05.Wk22.md
```
I use QuickAdd to add tasks to each day of the week.

The problem is that some weeks overlap into other months. Today is June 1st, so using QuickAdd to add to the file `gggg.MM.[Wk]w.md` doesn't add to `2023.05.Wk22.md >> # 6.01 Thursday` because we're now in June. But my _week_ is still in the last week of May.

image

Periodic Notes {{DATE:}} syntax automatically takes this into account when using Weekly Notes — my "Open Weekly Note" still opens `2023.05.Wk22.md` despite today being 6/01. However, I expect that it knows to do this because that's specifically a _weekly_ note that it's trying to open. 

QuickAdd is used for a lot more, so there are definitely people using it who would _want_ the current behavior of writing the current month `06` when they put `{{date:MM}}` on 6/01. However, I'd like some way to get "May" until next week when we're fully into "June".

Complicating this, I actually want to get a task added to the file `2023.05.Wk22.md` under the heading `6.01 Thursday`, which means the same QuickAdd capture would need to be able to add to this week's "month" filename, but use this (actual) "month" date in the heading.
```
2023.05.Wk22.md
# 5.29 Monday
# 5.30 Tuesday
# 5.31 Wednesday
# 6.01 Thursday
  ...                   ← (Want to add my task here)
# 6.02 Friday
```

I'm thinking a new token is needed for {{DATE:}} but I'd welcome any other ideas. Thanks!

chhoumann:
Closes #511.

## Problem

A weekly planner named `gggg.MM.[Wk]w` should file the week of **Thursday 2023-06-01** under `2023.05.Wk22.md` (the week belongs to May), but `{{DATE:gggg.MM.[Wk]w}}` renders `2023.06.Wk22` — moment has **no month-of-week token**, and QuickAdd passes the format slot to a single bare `window.moment()`. The reporter also wants the in-note heading (`# 6.01 Thursday`) to keep the actual day, so one capture needs two date contexts.

## Solution — a generic date *property*, not a new token

Instead of a `{{WEEK}}` clone, this adds a composable capability to the existing date tokens: **snap a date to the start/end of a period** before formatting, via pipe options consistent with `{{VALUE|...}}` / `{{VDATE|...}}` / `{{FILE|...}}`:

```
{{DATE:<format>|startof:<unit>}}     {{DATE:<format>|endof:<unit>}}
{{VDATE:<name>,<format>|startof:<unit>}}   (composes with |default |optional |time)
```

`<unit>` ∈ `year · quarter · month · week · isoweek · day` (case-insensitive). The locale-vs-ISO week choice is an **explicit unit** (`week` = locale first-day, `isoweek` = Monday) — no implicit anchor guessing. The `+N` day offset is unchanged and applied **before** the snap.

**#511 recipe (single picked date, dual context):**
`{{VDATE:d,gggg.MM.[Wk]w|startof:week}}` → `2023.05.Wk22` (filename) and `{{VDATE:d,M.DD dddd}}` → `6.01 Thursday` (heading).

## Why this shape

- **No new token / no new call site / no new preview code** — extends the existing `{{DATE}}`/`{{VDATE}}` passes already wired into all 5 formatter entry points (runtime ×2, both display previews, preflight).
- **Backward compatible** — the formatted-DATE regex peels only a trailing keyword-anchored `|startof:`/`|endof:` segment (lazy format slot), so every other literal `|` still renders verbatim: `{{DATE:YYYY|MM}}` → `2023|06`. Verified there are zero literal-pipe DATE formats in the repo.
- **VDATE per-occurrence snap** — snap-extraction is folded into `parseVDateOptions`, so all four call sites strip it from the default value automatically; the snap is applied at format time on the raw stored `@date:ISO`, so the same variable can render snapped in one place and day-actual in another.
- **Loud on typos** — an unknown unit throws a self-correcting error (moment's `startOf` silently no-ops otherwise); a mistyped key (`|starof:`) is left literal, byte-identical to today.

## Surfaces touched

`src/utils/dateModifiers.ts` (new shared kernel), `dates.ts` (`getDate` snap), `constants.ts` (regexes + syntax lists), `vdateSyntax.ts`, `formatters/formatter.ts`, `InsertAfterFields.svelte` (ordered-log date auto-detect tolerates the option), docs (`FormatSyntax.md`, Next only).

## Process

Designed, then reviewed with an ultracode multi-lens pass **plus two opposing-model (Codex) adversarial reviewers** on the design before implementation; their must-fixes (the broken `[|]` escape claim, the VDATE preview overrides, the four `parseVDateOptions` call sites, the test seam, a wrong offset example) are all folded in. `moment@2.29.4` added as a **dev-only** dependency (matches obsidian's pin) so the snap math is unit-tested against real moment; production still reads `window.moment`.

## Testing

- New unit tests: `dateModifiers`, `dates.snap` (the deterministic #511 guard + year-boundary + units + offset-then-snap + regex grammar incl. backward-compat & no-ReDoS), `vdateSyntax.snap` (snap extraction + legacy round-trip preservation). Full suite: **2458 passed, 0 failures**.
- `pnpm run build` (tsc + esbuild) and `pnpm run lint` green; `docs` build green; `pnpm install --frozen-lockfile` clean.
- **End-to-end verified in an isolated Obsidian vault** for both `{{DATE}}` and `{{VDATE}}`: week/month/quarter/isoweek snaps, `endof:month`, offset+snap, the dual-context recipe (`2023.05.Wk22 :: 6.01 Thursday`), `{{DATE:YYYY|MM}}` → `2023|06`, and bad-unit errors.

## Release Notes

* **New Features**
  * Added date “snap” support to template tokens via `|startof:<unit>` and `|endof:<unit>` for `{{DATE}}` and `{{VDATE}}`, applying snapping after any `+/-<N>` offset (with locale vs ISO week behavior).

* **Documentation**
  * Documented the syntax, supported units, precedence/ordering rules, and edge cases (including reserved pipe handling).

* **Bug Fixes**
  * Improved “insert after” date-format detection by ignoring snap options while preserving literal pipe characters.

* **Tests**
  * Expanded automated coverage for snapping, token parsing, grammar/performance, and deterministic behavior.

### DIFF
--- a/docs/docs/FormatSyntax.md
+++ b/docs/docs/FormatSyntax.md
@@ -14,6 +14,49 @@ Replace `<DATEFORMAT>` with a [Moment.js date format](https://momentjs.com/docs/
 
 Example: `{{DATE:YYYY-MM-DD_HH-mm}}` or `{{DATE:YYYY-MM-DD+3}}`.
 
+## `{{DATE:<DATEFORMAT>|startof:<unit>}}` / `{{...|endof:<unit>}}` {#date-snap}
+
+Snap the date to the **start** or **end** of a period before formatting. The
+formatted output then reflects that boundary instead of the exact instant — so,
+for example, the month of a week-snapped date is the month the *week* belongs to,
+not today's calendar month.
+
+`<unit>` is one of: `year`, `quarter`, `month`, `week`, `isoweek`, `day` (case-insensitive).
+
+- `week` uses your locale's first day of the week (matching the `w`/`ww`/`gggg` tokens).
+- `isoweek` uses Monday (matching the `W`/`WW`/`GGGG` tokens).
+
+| Token (on Thursday 2023-06-01) | Output |
+| --- | --- |
+| `{{DATE:gggg.MM.[Wk]w\|startof:week}}` | `2023.05.Wk22` |
+| `{{DATE:YYYY-MM\|startof:month}}` | `2023-06` |
+| `{{DATE:YYYY-MM-DD\|endof:month}}` | `2023-06-30` |
+| `{{DATE:YYYY-[Q]Q\|startof:quarter}}` | `2023-Q2` |
+| `{{DATE:GGGG-[W]WW\|startof:isoweek}}` | ISO week, Monday-anchored |
+
+**Weekly notes that cross months (issue #511).** A planner named `gggg.MM.[Wk]w`
+should file the week of June 1 under May (`2023.05.Wk22`), while the in-note
+heading still uses the actual day. Snap only the filename:
+
+```markdown
+# filename
+{{DATE:gggg.MM.[Wk]w|startof:week}}
+# heading inside the note
+{{DATE:M.DD dddd}}
+```
+
+This also works on `{{VDATE}}` — a single picked date can be week-snapped in one
+place and day-actual in another: `{{VDATE:d,gggg.MM.[Wk]w|startof:week}}` and
+`{{VDATE:d,M.DD dddd}}` share the same prompt. Combine freely with `|default`,
+`|optional`, and `|time` in any order.
+
+**Notes:**
+
+- The `+N` day offset is applied **before** the snap, so `{{DATE:YYYY-MM-DD+7|startof:week}}` is "the start of next week".
+- `endof:` snaps to the last moment of the period (`23:59:59.999`), so with a time format `{{DATE:YYYY-MM-DD HH:mm|endof:day}}` renders `... 23:59`.
+- `|startof:` and `|endof:` are the only reserved pipe options in a date format; any other literal `|` is still rendered verbatim (e.g. `{{DATE:YYYY|MM}}` → `2023|06`).
+- An unknown unit (e.g. `|startof:fortnight`) reports an error listing the valid units.
+
 ## `{{VDATE:<variable name>, <date format>}}` {#vdate}
 
 You'll get prompted to enter a date and it'll be parsed to the given date format. You could write 'today' or 'in two weeks' and it'll give you the date for that. Short aliases like `t` (today), `tm` (tomorrow), and `yd` (yesterday) are also supported and configurable in settings. Works like variables, so you can use the date in multiple places with different formats - enter once, format many times!
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "eslint-plugin-svelte": "^3.19.0",
     "globals": "17",
     "jsdom": "29",
+    "moment": "2.29.4",
     "obsidian": "1.13.1",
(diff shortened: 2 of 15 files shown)


chhoumann:
Thanks bots 🤖 — addressed in `f85d0d37` + `c1a5d4ea`, after a full ultracode + 2-Codex adversarial review of the implementation:

- **`detectDateFormatFromAfter` literal-pipe truncation** (CodeRabbit + Codex): fixed — extracted to `src/utils/insertAfterDateFormat.ts` with a keyword-anchored regex that preserves literal pipes (`{{DATE:YYYY|MM}}` → `YYYY|MM`), now table-tested.
- **Test `moment.locale("en")` global leak** (CodeRabbit): fixed — both snap suites restore the previous locale in `afterAll`.
- **ReDoS** (review-found, self-inflicted): the lazy format slot overlapped the snap alternation → O(n²) on a malformed unclosed token (n=30000 → ~3.2s). Replaced with a mutually-exclusive format arm + letter-anchored snap → linear (~1ms). The previous ReDoS test was a no-op (never entered the alternation) and is now a real `|startof:`×N assertion.
- **Bracket-literal abort** (Codex skeptic): `{{DATE:[label |startof: x ]YYYY-MM-DD|startof:month}}` used to peel the *first* `|startof:` and throw "Unknown date unit", aborting the whole format. Now only a trailing `|startof:<letters>` is the snap, so it renders correctly.
- **Untested wiring**: new `formatter-datesnap.test.ts` exercises the snap through `replaceDateInString`/`replaceDateVariableInString` with real moment + a frozen clock (mutating the capture group used to keep the suite green).

Known limitation (accepted): the VDATE *builder preview* shows the un-snapped date (the preview uses a hand-rolled, moment-less generator that already can't render `gggg`); runtime output is correct.

chhoumann:
Latest CodeRabbit nitpick (multi-token heading coverage) addressed in `52595c20` — added cases for a multi-token heading with no date token (→ `undefined`) and one with a date token (→ the date format). Also merged `master` (resolved the `InsertAfterFields` import conflict from #1370). All required checks green, CodeRabbit ✅, `MERGEABLE/CLEAN`; full suite 2500 passed; feature + ReDoS/bracket fixes re-verified end-to-end in an isolated vault.

PR #1375, opened 2026-06-17
fix(format): add trim option for value tokens

### LINKED ISSUE #685: [FEATURE REQUEST] Provide way to capture value and trim leading and trailing whitespace

When using the capture to create a new note sometimes a trailing space is left in the input (especially on mobile). This breaks the linking that is created when appending the link until I rename the file. 

Provide a an alternative to {{VALUE}} that will trim the leading and trailing whitespace to the input {{TRIMEDVALUE}}  as an example

**Additional context**

chhoumann:
Add an explicit `|trim` VALUE option so users can render captured input without leading/trailing whitespace when that value is used in note names, links, or other whitespace-sensitive locations.

The underlying problem in #685 is not that QuickAdd needs another top-level placeholder; it is that mobile/keyboard input can accidentally include surrounding spaces, and using raw `{{VALUE}}` in a capture target can create a different note path than intended. This keeps `{{VALUE}}` fully backwards-compatible and makes trimming opt-in per token:

- `{{VALUE|trim}}` / `{{NAME|trim}}`
- `{{VALUE:title|trim}}`
- composition with existing transforms, e.g. `{{VALUE:title|trim|case:slug}}`

Alternatives considered:

- `{{TRIMMEDVALUE}}`: rejected because it duplicates the VALUE grammar and does not generalize to named VALUE tokens or existing pipe options.
- Auto-trimming all capture values: rejected because users may intentionally capture surrounding whitespace or want raw and trimmed forms in the same format.
- Capture-only setting: rejected because the formatter already has per-token transform syntax and this is useful anywhere VALUE syntax is used.

Implementation notes:

- `|trim` is a render transform only; cached/stored variables remain raw, so `{{VALUE:title}}` and `{{VALUE:title|trim}}` can be used together.
- Trim is applied before `|case:` and before YAML property collection sees transformed string values.
- `|trim:false` is accepted for shared-snippet override consistency.
- `|trim` is ignored with `|multi` arrays, matching the existing `|case` guard.
- Because `trim` is now a reserved bare flag, a literal default value of `trim` should use `|default:trim`.

Live Obsidian validation, isolated worktree vault:

- Built and provisioned with `pnpm run build` and `pnpm run provision:e2e-vault`.
- Reproduced current behavior: capture target `__qa_685/plain/{{VALUE}}` with `value="  Mobile Title  "` created `__qa_685/plain/  Mobile Title.md` and body `Plain:   Mobile Title  `.
- Verified fix: capture target `__qa_685/trim/{{VALUE|trim}}` with the same value created `__qa_685/trim/Mobile Title.md` while body output showed raw `{{VALUE}}`, trimmed `{{VALUE|trim}}`, and `{{VALUE|trim|case:slug}}` as expected.
- Verified named value path: `{{VALUE:title|trim}}` created `__qa_685/named/Named Title.md` while raw `{{VALUE:title}}` still preserved spaces.
- `pnpm run obsidian:e2e -- dev:errors` reported no captured runtime errors.

Validation:

- `pnpm run build`
- `pnpm run provision:e2e-vault`
- `pnpm run obsidian:e2e -- plugin:reload id=quickadd`
- `pnpm run obsidian:e2e -- quickadd:run ...` for plain, trimmed, and named trimmed capture choices
- `pnpm run obsidian:e2e -- dev:errors`
- `pnpm run lint`
- `pnpm run check` (0 errors, existing 4 warnings)
- `pnpm test` (225 files passed, 5 skipped; 2648 tests passed, 37 skipped)
- `pnpm run build-with-lint`

Fixes #685

## Summary by CodeRabbit

* **New Features**
  * Added a `trim` modifier for `VALUE` and variable tokens (including file-name contexts) to strip leading/trailing whitespace (e.g., `{{VALUE|trim}}`, `{{VALUE:<name>|trim}}`).
  * `trim` is applied per token and no longer affects stored underlying values.
  * `trim` composes with case transformations and runs before `case` (e.g., `trim|case:slug`).

* **Documentation**
  * Updated format syntax docs to clarify default handling rules and the `trim`/`optional` reserved-flag behavior, plus `multi` compatibility notes.

* **Tests / Bug Fixes**
  * Improved parsing and added coverage to ensure `trim` isn’t treated as a legacy default and behaves correctly across token types and non-string values.

### DIFF
--- a/docs/docs/FormatSyntax.md
+++ b/docs/docs/FormatSyntax.md
@@ -143,7 +143,7 @@ Example: `priority: {{VALUE:🔽,🔼,⏫|text:Low,Normal,High}}`.
 
 ## `{{VALUE:<variable name>|<default>}}` {#value-default}
 
-Same as above, but with a default value. For single-value prompts (e.g., `{{VALUE:name|Anonymous}}`), the default is pre-populated in the input field - press Enter to accept or clear/edit it. For multi-value suggesters without `|custom`, you must select one of the provided options (no default applies). If you combine keyed options like `|label:`, `|default:`, `|type:`, or `|case:`, shorthand defaults like `|Anonymous` are ignored; use `|default:Anonymous` instead. The bare `|optional` flag is the exception: `{{VALUE:name|Anonymous|optional}}` keeps the shorthand default. Because `optional` is now a reserved flag word, a literal default of "optional" needs the keyed form: `|default:optional`.
+Same as above, but with a default value. For single-value prompts (e.g., `{{VALUE:name|Anonymous}}`), the default is pre-populated in the input field - press Enter to accept or clear/edit it. For multi-value suggesters without `|custom`, you must select one of the provided options (no default applies). If you combine keyed options like `|label:`, `|default:`, `|type:`, `|case:`, or `|trim:`, shorthand defaults like `|Anonymous` are ignored; use `|default:Anonymous` instead. The bare `|optional` flag is the exception: `{{VALUE:name|Anonymous|optional}}` keeps the shorthand default. Because `optional` and `trim` are now reserved flag words, literal defaults of "optional" or "trim" need the keyed form, e.g. `|default:trim`.
 
 Example: `status: {{VALUE:status|Draft}}`.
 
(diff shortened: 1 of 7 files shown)


chhoumann:

PR #1376, opened 2026-06-17
feat: support clipboard images in captures

### LINKED ISSUE #1161: [FEATURE REQUEST] {{clipboard}} to support image type
Currently `{{clipboard}}`only supports the text type of clipboard content.

It would be helpful if `{{clipboard}}` would also support the image type of the clipboard content. So we can insert pictures.

**Additional context**
Perhaps formatted RTF also could be hands, but only as an option.

chhoumann:
Add image fallback support for `{{clipboard}}` in Capture choice content.

The underlying problem is that users can capture text from the clipboard but cannot capture an image-only clipboard into an Obsidian note. This keeps the existing token instead of adding a one-off image token because the user intent is still “insert clipboard contents.” To preserve compatibility, clipboard text continues to win; only when the clipboard text is empty does QuickAdd look for a supported image item.

Design notes:
- Scoped to `CaptureChoiceFormatter`, so templates, file names, previews, API formatting, and macro formatting remain text-only and side-effect free.
- Uses Obsidian’s `getAvailablePathForAttachment()` and `vault.createBinary()` so attachment folder settings and filename de-duping stay native.
- Reuses one saved attachment link for repeated `{{clipboard}}` tokens in a single capture run.
- Makes clipboard token replacement single-pass so clipboard text containing `{{clipboard}}` is inserted literally instead of being re-expanded.

Validation:
- Reproduced issue in this worktree’s isolated Obsidian vault: image-only clipboard produced `Copied: ` and no attachment.
- Verified fix in the same isolated vault: image-only clipboard produced `Copied: ![[Clipboard image 2026-06-17 20.35.48.png]]` and created the PNG attachment.
- Verified text clipboard still produces `Copied: plain text after fix` and does not read image data when text exists.
- `pnpm run obsidian:e2e -- dev:errors` -> no errors captured.
- `pnpm run check` -> 0 errors, existing 4 warnings.
- `pnpm test` -> 2643 passed, 37 skipped.
- `pnpm run build-with-lint` -> passed.

Fixes #1161

## Release Notes

* **New Features**
  * Clipboard images are now supported. When using the `{{CLIPBOARD}}` token with an image in your clipboard, it is automatically saved as an attachment and embedded in your note. Text content takes precedence if both are available.

* **Documentation**
  * Updated `{{CLIPBOARD}}` token documentation to reflect image capture behavior.

### DIFF
--- a/docs/docs/FormatSyntax.md
+++ b/docs/docs/FormatSyntax.md
@@ -415,6 +415,8 @@ Example: `> {{selected}}`.
 
 The current clipboard content. Will be empty if clipboard access fails due to permissions or security restrictions.
 
+In Capture choices, if the clipboard has no text but contains an image, QuickAdd saves the image using Obsidian's attachment location settings and inserts an embedded attachment link. If the clipboard has text, the text is used.
+
 Example: `Copied: {{CLIPBOARD}}`.
 
 ## `{{RANDOM:<length>}}` {#random}
--- a/src/formatters/captureChoiceFormatter-clipboard.test.ts
+++ b/src/formatters/captureChoiceFormatter-clipboard.test.ts
@@ -1,0 +1,220 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { App, TFile } from "obsidian";
+
+vi.mock("../utilityObsidian", () => ({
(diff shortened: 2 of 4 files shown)


chhoumann:

PR #1377, opened 2026-06-17
fix: sanitize generated note paths

### LINKED ISSUE #221: Quick Add will make new note unlinkable if hit enter / return
**Describe the bug**
When I use Quick Add, it pops up a window for me to fill in the {{VALUE}}. If I hit return / enter, the new note will be unlinkable. 

**To Reproduce**
1. Using Quick Add to add new note
2. Fill in any VALUE
3. Hit Enter / Return. Then click OK.
4. From another note, link to the new note (Created by Quick Add) heading or block. It will not link to that new note!!

**Expected behavior**
It should link to the new note created by Quick Add.

**Screenshots**
- Create New note using Quick Add. (I can hit return to second line without filling anything and it would create the same error)
2022-03-25 at 10 42 57 PM (Obsidian)

- Cannot link to the New note created Quick Add
2022-03-25 at 10 43 34 PM (Obsidian)

 - MacOS Monterey Version 12.3
 - Browser Safari
 - Version 15.4
- Obsidian v0.14.2

chhoumann:
Fixes #221

QuickAdd could create notes whose generated file name contained hidden newline/control characters when a value used in a Template or Capture destination came from a multi-line prompt or another variable source. Obsidian can create those files, but links and heading/block resolution become unreliable because the visible note title does not match the actual path.

This fixes the invariant at the file-path boundary instead of changing VALUE prompt semantics: generated Template and Capture note paths now collapse C0 control-character runs inside path segments before checking collisions or creating files. Normal path separators are still preserved, and explicit `.md` / `.canvas` capture suffixes are handled after normalizing the generated stem so `{{VALUE}}.md` with a trailing newline becomes `Value.md`, not `Value .md`.

I considered making Enter submit for all filename prompts, but single-line prompts already do that and multi-line VALUE prompts are valid for note bodies and captures. Path normalization protects CLI, URI, script, global-variable, and prompt-driven inputs consistently without narrowing `{{VALUE}}` globally.

Validation evidence:
- Reproduced current bug in this worktree's isolated Obsidian vault: newline input created `qa-issue-221-before/This is the VALUE\n.md`.
- Verified fix live after reload: the same Template input created `qa-issue-221-after/This is the VALUE.md` with no runtime errors.
- Verified final bundle live: Template `Template {{VALUE}}` and Capture `Capture {{VALUE}}.md` both created normal `Name.md` paths from `Name\n`; `dev:errors` reported no errors.
- `pnpm run lint`
- `pnpm run check` (0 errors, existing 4 warnings)
- `pnpm test`
- `pnpm run build-with-lint`

No settings migration or user-facing docs change is needed; this only prevents QuickAdd from committing invalid hidden-control-character note paths.

### DIFF
--- a/src/engine/CaptureChoiceEngine.selection.test.ts
+++ b/src/engine/CaptureChoiceEngine.selection.test.ts
@@ -483,6 +483,32 @@ describe("CaptureChoiceEngine capture target resolution", () => {
 		expect(resolved).toBe("Inbox/note.md");
 	});
 
+	it("collapses newlines in generated capture target file names", async () => {
+		const engine = new CaptureChoiceEngine(
+			createApp(),
+			{ settings: { useSelectionAsCaptureValue: false } } as any,
+			createChoice({ captureTo: "Inbox/This is the VALUE\n" }),
+			createExecutor(),
+		);
+
+		const resolved = await (engine as any).getFormattedPathToCaptureTo(false);
+
+		expect(resolved).toBe("Inbox/This is the VALUE.md");
+	});
+
+	it("collapses newlines before an explicit capture target extension", async () => {
+		const engine = new CaptureChoiceEngine(
+			createApp(),
+			{ settings: { useSelectionAsCaptureValue: false } } as any,
+			createChoice({ captureTo: "Inbox/This is the VALUE\n.md" }),
+			createExecutor(),
+		);
+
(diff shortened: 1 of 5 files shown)


chhoumann:
Closing: mis-launched settings; relaunched cleanly.

PR #1378, opened 2026-06-17
fix(capture): keep cursor at active-file capture position

### LINKED ISSUE #348: [FEATURE REQUEST] Keep cursor at position of capture
Currently, I use quick to add to capture a work log, thoughts, or other things i would like to save quickly. 
It would be great if we can not only open the file we add the capture to but also have the cursor stay there.

The reason is that sometimes i just create a note. Sometimes i want to add a link to a another note next to it. and now I need my mouse for that.

chhoumann:
Fixes #348.

Active-file Capture insertions currently leave the editor cursor after the inserted capture text. That breaks the quick writing flow for users who capture a snippet and then want to keep typing or add a link at the original position without reaching for the mouse.

This keeps the change scoped to Capture modes that write through the active Markdown editor:

- current-line / at-cursor Capture restores the original collapsed cursor after inserting the capture text
- new-line-above Capture restores the cursor with the original line after the inserted block shifts it down
- new-line-below Capture keeps the cursor in its original position

I did not add a setting. The behavior is a bug fix for an existing placement mode rather than a new convention, and keeping the cursor at the capture position matches the user intent of “capture here, then continue here.” File append/prepend, insert-after/before, Canvas, open-file behavior, and link insertion behavior are left unchanged. Non-collapsed selections keep the editor default replacement behavior so this does not collapse or discard selection state for selection-based captures.

Implementation notes:

- Cursor restoration uses pre-insert editor offsets, then maps back to a position after mutation. That matters for multiline insertions, especially inserting above a line when the first inserted line is shorter than the cursor column.
- Unit tests cover current-line, above, below, multiline inserted content, and selected-text replacement behavior.
- E2E coverage drives real Obsidian through the isolated worktree vault for all three active-file Capture modes.

Validation:

- Reproduced current behavior in the isolated Obsidian vault before the fix: inserting `CAPTURED ` into `alpha beta` at cursor `0:6` produced `alpha CAPTURED beta` and moved the cursor to `0:15`.
- Verified after the fix in the isolated Obsidian vault: the same capture produces `alpha CAPTURED beta` and leaves the cursor at `0:6` with no captured runtime errors.
- `pnpm run build-with-lint`
- `pnpm run check` (0 errors, existing warnings only)
- `pnpm test`
- `eval "$(pnpm run start:e2e-obsidian -- --print-env | sed -n '/^export QUICKADD_E2E_/p')" && pnpm exec vitest run --config vitest.e2e.config.mts tests/e2e/capture-cursor-position.test.ts`

Docs: not updated; this is a behavior fix for existing Capture placement modes, with no new setting, syntax, or migration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced cursor position preservation when inserting captured content at the current line, above, or below in the editor.

* **Tests**
  * Added comprehensive unit tests for cursor and selection preservation during text insertion operations.
  * Added end-to-end test suite validating that QuickAdd Capture maintains stable editor cursor and selection positions across various scenarios.

### DIFF
--- a/src/utils/editorInsertion.test.ts
+++ b/src/utils/editorInsertion.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, it } from "vitest";
+import type { App, EditorPosition } from "obsidian";
+import { appendToCurrentLine, insertOnNewLine } from "./editorInsertion";
+
+class TextEditor {
+	private value: string;
+	private cursor: EditorPosition;
+
+	constructor(value: string, cursor: EditorPosition) {
+		this.value = value;
+		this.cursor = { ...cursor };
+	}
+
+	getValue(): string {
+		return this.value;
+	}
+
+	getCursor(): EditorPosition {
+		return { ...this.cursor };
+	}
+
+	setCursor(position: EditorPosition): void {
+		this.cursor = { ...position };
+	}
+
+	getLine(line: number): string {
+		return this.value.split("\n")[line] ?? "";
+	}
+
+	posToOffset(position: EditorPosition): number {
+		const lines = this.value.split("\n");
+		let offset = 0;
+		for (let line = 0; line < position.line; line++) {
+			offset += (lines[line]?.length ?? 0) + 1;
+		}
+		return offset + position.ch;
+	}
+
+	offsetToPos(offset: number): EditorPosition {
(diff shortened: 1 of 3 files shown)


chhoumann:

PR #1379, opened 2026-06-17
feat(capture): support file-backed capture formats

### LINKED ISSUE #614: [FEATURE REQUEST] Capture format can refer to a file

Capture format is powerful but very difficult to edit. I had a couple script in capture format that is very long. It would be nice if I can just refer to file in the vault or template. Then I can just edit the file there instead. Beside, I want portability where I can view the script outside of obsidian.

Under capture format toggle, we add another option. 
*Use template file* toggle. When it turn on, we change it to file selector. The content within the file will get execute just like the text within existing behavior.

**Additional context**
I tried to use template, but it doesn't have the same function.

chhoumann:
Fixes #614

Capture format already had the right runtime primitive for this request: `{{TEMPLATE:path}}` can render a vault file through the same formatter pipeline as inline capture text. This PR makes that capability complete and discoverable instead of adding a second persisted “use template file” mode that would duplicate format syntax and complicate choice settings.

What changed:
- Documents the Capture-specific workflow for keeping long capture bodies in vault files with `{{TEMPLATE:Templates/Capture Body.md}}`.
- Updates the Capture form description so users can discover the TEMPLATE token from the setting itself.
- Preflights `{{TEMPLATE:...}}` files referenced from Capture format, so one-page input / CLI collection sees prompts inside the file-backed body.
- Includes Capture-format template files in package export and import preview, including nested `{{TEMPLATE:...}}` includes discovered inside bundled template assets.

Design tradeoffs:
- I did not add a new `Use template file` setting. The existing format token is more general: it works inline, can include multiple files, can nest, and preserves all Capture-specific positioning/target behavior.
- `{{TEMPLATE:...}}` still requires an explicit `.md`, `.canvas`, or `.base` extension because that is the current formatter contract. The docs now say this directly rather than broadening path resolution in a package/discoverability fix.

Validation:
- Reproduced the underlying issue in the current product shape: long Capture bodies can only be discovered as generic format syntax, and package traversal did not treat Capture-format template references as package assets.
- Live Obsidian validation in this worktree’s isolated vault:
  - `pnpm run build`
  - `pnpm run provision:e2e-vault`
  - Seeded `Templates/Capture Body.md` with `File-backed body: {{VALUE:topic}}`
  - Seeded a Capture choice whose format is `{{TEMPLATE:Templates/Capture Body.md}}`
  - Ran `pnpm run obsidian:e2e -- quickadd:run choice="Issue 614 Capture" value-topic="validated from file"`
  - Verified `Issue614 Target.md` contained `File-backed body: validated from file`
  - Verified `pnpm run obsidian:e2e -- dev:errors` reported `No errors captured`
- Automated gates:
  - `pnpm exec vitest run src/services/packageExportService.test.ts src/services/packagePreview.test.ts src/preflight/collectChoiceRequirements.test.ts src/utils/packageTraversal.test.ts`
  - `pnpm run check` (0 errors; existing warnings remain)
  - `pnpm test`
  - `pnpm run build-with-lint`

Notes for reviewers:
- Package export discovers nested template includes while reading template assets, not in pure choice traversal, because nested include discovery requires file contents.
- Import preview expands nested references only from bundled template assets; unbundled/missing template files cannot be inspected until they exist in the vault.

### DIFF
--- a/docs/docs/Choices/CaptureChoice.md
+++ b/docs/docs/Choices/CaptureChoice.md
@@ -429,6 +429,18 @@ If you do not enable this, QuickAdd will default to `{{VALUE}}`, which will inse
 
 You can use [format syntax](/FormatSyntax.md) here, which allows you to use dynamic values in your capture format.
 
+Long capture formats can live in a normal vault file. Put the body in a
+template file, then reference it from **Capture format**:
+
+```md
+{{TEMPLATE:Templates/Capture Body.md}}
+```
+
+QuickAdd renders that file as the capture body, so the file can still contain
+normal format syntax, inline `js quickadd` blocks, macros, and nested template
+includes. Use an explicit `.md`, `.canvas`, or `.base` extension in the
+`{{TEMPLATE:...}}` path.
+
 If you want to insert `.base` content into your current note, keep **Capture to active file** enabled and use a `.base` template token in the capture format. See [Capture: Insert a Related Notes Base into an MOC Note](/Examples/Capture_InsertBaseTemplateIntoActiveFile.md).
 If you want QuickAdd to create a brand new note that already contains an
 embedded Base, use a Template choice instead. See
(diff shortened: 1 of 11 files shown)


chhoumann:

PR #1382, opened 2026-06-17
feat: copy produced note links to clipboard

### LINKED ISSUE #600: [FEATURE REQUEST] Add option to copy the link to a newly created note through a template action.
It would be nice if we had the option to copy the link to the clipboard alongside the option to insert it in the active note. Sometimes I don't want to insert the link to the note that is active, but in another note and having the link in the clipboard could be helpful.

chhoumann:
Adds a copy-to-clipboard link action for Template and Capture choices, so users can create or capture to a note and paste that note link somewhere else without requiring insertion into the active note.

The underlying problem in #600 is not just Template insertion placement: the produced note link needs to be portable. I extended the existing `appendLink` model with `copyToClipboard` instead of adding a Template-only setting, so Template and Capture choices share the same Linking section and can be configured as insert-only, copy-only, both, or neither. The copied text is generated through Obsidian's `fileManager.generateMarkdownLink`, matching the existing inserted-link behavior and respecting vault link settings.

Compatibility notes:
- Existing boolean `appendLink` settings still normalize to copy disabled, so no migration is required.
- Clipboard write failures are treated as non-fatal post-commit side effects: the note/capture remains successful and QuickAdd shows/logs the copy failure.
- This applies to configured Template/Capture choices. The ad hoc "New note from template" launcher does not get a new global copy default in this PR.

Validation:
- Reproduced baseline in the isolated worktree vault: Template choice created `qa-600-created.md` while clipboard stayed `BASELINE_CLIPBOARD`.
- Verified fix in the isolated worktree vault: Template copy-only choice created `qa-600-manual-created.md` and clipboard became `[[qa-600-manual-created]]`.
- Verified Capture fix in the isolated worktree vault: Capture copy-only choice wrote `qa-600-manual-capture-created.md` and clipboard became `[[qa-600-manual-capture-created]]`.
- `pnpm run check`
- `pnpm run test`
- `pnpm run build-with-lint`
- `QUICKADD_E2E_* ... pnpm exec vitest run --config vitest.e2e.config.mts tests/e2e/copy-link-to-clipboard.test.ts`

Fixes #600

## Summary by CodeRabbit

* **New Features**
  * Added a "Copy link to clipboard" toggle for Capture and Template choices, enabling users to automatically copy created file links to the clipboard without inserting them into the active note. This option operates independently from link insertion settings.

### DIFF
--- a/docs/docs/Choices/CaptureChoice.md
+++ b/docs/docs/Choices/CaptureChoice.md
@@ -130,6 +130,8 @@ The Capture builder is grouped into sections: **Location**, **Position**, **Link
     -   **Enabled (skip if no active file)** – inserts the link when possible and silently drops `{{LINKCURRENT}}` if nothing is open
     -   **Disabled** – never append a link
 
+    Enable _Copy link to clipboard_ to copy the captured file's Obsidian link after the Capture choice finishes. This can be used by itself, so the link is available for pasting somewhere else without inserting it into the active note.
+
     When either enabled mode is selected, a _Link placement_ dropdown appears so you can choose where the link is placed:
     -   **Replace selection** - Replaces any selected text with the link (default)
     -   **After selection** - Preserves selected text and places the link after it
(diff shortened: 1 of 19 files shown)


chhoumann:

PR #1428, opened 2026-06-27
fix(cli): abort non-interactive runs instead of hanging on engine prompts

chhoumann:
## Summary

Found via a live end-to-end audit of QuickAdd in an isolated Obsidian instance: a **non-interactive CLI run** (`quickadd:run` / `quickadd:run-template` without the `ui` flag, and any headless automation) **hangs forever** when an engine opens a blocking prompt the requirement collector cannot pre-satisfy.

### Reproduction (before this PR)
```
# 1st run creates the note
obsidian vault=… quickadd:run-template path=Templates/T.md value-value=Daily
# 2nd run — target now exists, default "Ask every time" file-exists behaviour:
obsidian vault=… quickadd:run-template path=Templates/T.md value-value=Daily
# → no output, no error envelope, hangs 2+ minutes (stuck on an unanswerable
#   "file already exists" suggester). Same for quickadd:run of any Template
#   choice whose fileExistsBehavior is "prompt".
```

The non-interactive CLI contract is "succeed, or return an `{ok:false}` envelope" — never hang. The requirement collector pre-collects the inputs it can *see*, but prompts opened at **runtime** (after token resolution) escape it and block indefinitely.

## Reproduced hang sites (all fixed + verified live)
| Site | Trigger |
|------|---------|
| Template file-exists prompt | target note exists + `fileExistsBehavior: prompt` (the default) |
| Template folder chooser | `chooseWhenCreatingNote` / multiple folders |
| Capture target file picker | format-syntax `Capture to` (e.g. `{{VALUE:folder}}`) |
| Capture heading picker | `Choose heading when capturing` |
| Formatter token prompts | `{{VALUE}}` / `{{VDATE}}` / … hidden behind a tokenized template path or capture target |
| Macro member picker | a user script that exports an object of named members |
| AI model / prompt-template pickers | `Ask me` model, or prompt-template selection |

## Fix
- Add `interactive?: boolean` to `IChoiceExecutor` (default `true` → **every GUI entry point is byte-identical**). `ChoiceExecutor` defaults it `true`; the CLI sets it `false` when `ui` is absent.
- Each **config-driven** prompt the collector cannot pre-satisfy now throws an actionable `ChoiceAbortError` → a clean `ok:false` envelope in ~1–30 ms instead of hanging:
  - `TemplateChoiceEngine`: file-exists prompt + folder chooser (`getOrCreateFolder`)
  - `CaptureChoiceEngine`: capture target file picker + heading picker
  - `CompleteFormatter`: a single `assertInteractivePrompt` guard covering all 7 token-prompt methods (the central source of token prompts). `RequirementCollector` overrides `promptForValue` separately, so **input collection is unaffected**.
  - `MacroChoiceEngine` / `AIAssistant`: member picker, AI `Ask me` model picker, AI prompt-template picker
- The abort reason is threaded through `ChoiceOutcome.cancelled.reason` so `quickadd:run-template` surfaces the actionable message; the **URI x-callback handler still reports only `cancelKind`**, so no vault detail leaks to an external callback URL. A user dismissal keeps its stable "cancelled by user" text.

### Boundary (intentional, not a bug)
A macro user-script that **explicitly** calls an interactive api method (`api.inputPrompt`/`suggester`/`yesNoPrompt`/`checkboxPrompt`) is inherently interactive and still needs `ui`. That's the script author's choice, not a config-driven prompt, so the public api contract is left unchanged.

## Verification
- **Live** (this worktree's isolated Obsidian instance): every site above now returns a clear `{ok:false, error, aborted:true}` envelope in ~1–30 ms; new-note create / normal capture / normal macro are unaffected; no leaked modals.
  - e.g. `run-template` on existing target → `'T' needs to ask what to do because a note with that name already exists … Set the choice's "If the file already exists" behaviour … or re-run with the ui flag.`
- **Gates**: `tsc` ✓, `eslint` ✓, `svelte-check` 0 errors ✓, `pnpm test` 3169 passed (+ regression tests for the file-exists, folder, heading, and formatter guards).
- Two rounds of adversarial review (opposite model); the first round caught that an earlier version only guarded 3 sites — this PR extends the fix to the full reproduced surface.

## Summary by CodeRabbit

* **New Features**
  * Added a non-interactive mode that prevents prompt/picker/template selection UI from blocking headless runs.
  * Exposed an `interactive` setting to control whether interactive selection is allowed (including AI assistant template selection).

* **Bug Fixes**
  * Improved cancellation results: user cancels are distinguished from aborts, and aborted/cancelled responses can include a clearer reason when available.
  * CLI behavior now maps interactivity to whether the `ui` option is enabled.

* **Tests**
  * Added/extended coverage for non-interactive abort behavior across capture, template collisions, and formatting/assistant flows.

### DIFF
--- a/src/IChoiceExecutor.ts
+++ b/src/IChoiceExecutor.ts
@@ -29,6 +29,17 @@ export interface IChoiceExecutor {
 	): Promise<ChoiceOutcome>;
 	variables: Map<string, unknown>;
 	/**
+	 * Whether this execution may open blocking interactive UI (suggesters/modals)
+	 * for inputs the requirement collector cannot pre-satisfy — e.g. the
+	 * "file already exists" prompt, the folder chooser, or the heading picker.
+	 * Defaults to interactive (`true`/`undefined`) so GUI runs are unchanged.
+	 * Non-interactive callers (the `quickadd:run`/`run-template` CLI without the
+	 * `ui` flag, and any headless automation) set this to `false`; engines then
+	 * abort such a prompt with a clear, actionable error instead of hanging forever
+	 * on an unanswerable modal.
+	 */
+	interactive?: boolean;
+	/**
 	 * Frontmatter property value focused when the outermost choice execution began.
 	 * Append Link uses this to avoid the stale CodeMirror cursor Obsidian reports
 	 * while a Properties field owns focus.
--- a/src/ai/AIAssistant.ts
+++ b/src/ai/AIAssistant.ts
@@ -1,6 +1,7 @@
 import type { App } from "obsidian";
 import { TFile } from "obsidian";
 import { UserCancelError } from "src/errors/UserCancelError";
+import { ChoiceAbortError } from "src/errors/ChoiceAbortError";
 import GenericSuggester from "src/gui/GenericSuggester/genericSuggester";
 import { settingsStore } from "src/settingsStore";
 import { getMarkdownFilesInFolder } from "src/utilityObsidian";
@@ -176,7 +177,8 @@ function toError(reason: unknown): Error {
 async function getTargetPromptTemplate(
 	app: App,
 	userDefinedPromptTemplate: Params["promptTemplate"],
-	promptTemplates: TFile[]
+	promptTemplates: TFile[],
+	interactive = true
 ): Promise<[string, string]> {
 	let targetFile;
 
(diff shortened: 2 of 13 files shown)


chhoumann:
Addressed both CodeRabbit findings:
- **MacroChoiceEngine (Major):** single-member script exports now auto-run in non-interactive mode (only multi-member exports abort). Verified live: a macro whose script exports one member runs headlessly and creates its file; multi-member still returns a clean `ok:false` abort.
- **CaptureChoiceEngine (Minor):** moved `assertInteractiveCaptureTarget()` after the empty-folder / no-match `invariant` in both `selectFileInFolder` and `selectFileFromSet`, so a genuinely empty scope surfaces its accurate "Folder … is empty." error instead of the prompt message.

Gates re-run green (tsc, eslint, 3169 tests).

PR #1452, opened 2026-06-29
docs(examples): make migrate-dataview inline-code masker linear (ReDoS)

chhoumann:
## What

`maskInlineCode` in the `migrateDataviewToFrontmatter` example userscript masked inline code spans with a regex:

```js
text.replace(/(`+)(?:(?!\1)[\s\S])*?\1/g, (span) => ' '.repeat(span.length))
```

That combines a **backreferenced opener** `` (`+) ``, a **lazy body**, and a **negative lookahead** `(?!\1)`. On a long backtick run with no matching-length closer it backtracks super-linearly. The masker runs (L315) on **every** non-task body line containing `::`, before any allow-list filtering, and the script reads the active note via `app.vault.read`. So a single body line like `` x:: ``` `` followed by thousands of backticks - in a **synced / shared / imported** note the user opens and migrates - freezes Obsidian's main thread.

This is the **same ReDoS class** already removed from the in-bundle `InlineFieldParser` (PR #1444); this distributed example script kept its own vulnerable masker (introduced in PR #1439, untouched by #1447).

## Reproduction (red)

Timing the original masker on `` 'x:: ' + '`'.repeat(N) ``:

| N backticks | old regex |
|------------:|----------:|
| 16,000 | 85 ms |
| 60,000 | **1,163 ms** |

Growth is ~O(n²·¹); a 120-200 KB line freezes the UI thread for many seconds.

## Fix

Replace the regex with a **single linear pass over the backtick runs** that reproduces the regex's exact masking output in O(n): a run of `k` backticks is closed by the first run of **at least** `k` backticks (its first `k` form the closer, the regex's `\1`), preferring the longest opener length the engine would have matched; an unclosed opener stays verbatim. Output length is preserved exactly (positions stay aligned for the caller's slice logic).

## Proof (green)

- **Equivalence**: the new scanner is **byte-for-byte identical** to the old regex over **111.5M inputs** - exhaustive over `` {`, a, \r, \n, \t, emoji, :, combining-mark, space} `` up to length 8, exhaustive triple-run patterns, and millions of randomised structured inputs. Zero divergence, zero length drift.
- **Linearity**: `` 'x:: ' + '`'.repeat(N) `` - N=60,000 → **0.40 ms** (was 1,163 ms); N=1,000,000 (a 1 MB line) → **13.8 ms**.
- **End-to-end** through the real call site (`parseInlineFieldsWithWikilinks`) on a poisoned note with a 200k-backtick line: whole migration runs in **4.1 ms** (was a multi-second freeze), real `Reference`/`Related` fields still migrate, the fenced code block and frontmatter are preserved, and the pathological line is left verbatim. **#1439/#1447 fence + wikilink handling not regressed.**

## Scope / release

Example userscript under `docs/static/scripts` → `docs(examples):` (no `feat`/`fix`/`perf`, so no release is triggered). Tests are kept local (harness only), per repo convention for example scripts.

## Gates

`tsc -noEmit` ✅ · `eslint .` ✅ · `svelte-check` ✅ (0 errors) · `vitest` ✅ (3444 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of inline code text during migration, with more reliable masking and fewer edge-case formatting issues.
  * Preserved text spacing and unmatched backticks more consistently when processing content.

### DIFF
--- a/docs/static/scripts/migrateDataviewToFrontmatter.js
+++ b/docs/static/scripts/migrateDataviewToFrontmatter.js
@@ -470,9 +470,86 @@ function isClosingFence(line, fenceChar, fenceLen) {
  * mistaken for an inline-field separator. Supports arbitrary backtick-run lengths
  * (`code`, ``co`de``). Character positions are preserved so the caller can slice
  * the original line by offsets computed on the masked line.
+ *
+ * Implemented as a single linear pass over the backtick runs. The earlier regex
+ * /(`+)(?:(?!\1)[\s\S])*?\1/g combined a backreferenced opener, a lazy body, and a
+ * negative lookahead, which backtracks super-linearly (seconds of main-thread
+ * freeze) on a long backtick run with no matching-length closer - reachable from a
+ * synced/shared note that the user migrates. This scanner reproduces the regex's
+ * exact masking behaviour with O(n) work: a run of k backticks is closed by the
+ * first run of at least k backticks (its first k backticks form the closer, the
+ * regex's `\1`), preferring the longest opener length the engine would have
+ * matched. An unclosed opener stays verbatim.
  */
 function maskInlineCode(text) {
-    return text.replace(/(`+)(?:(?!\1)[\s\S])*?\1/g, (span) => ' '.repeat(span.length));
+    const n = text.length;
+
+    // Index every maximal run of backticks once.
+    const runStart = [];
+    const runLen = [];
+    for (let i = 0; i < n; ) {
+        if (text.charCodeAt(i) === 96 /* ` */) {
+            let j = i + 1;
+            while (j < n && text.charCodeAt(j) === 96) j++;
+            runStart.push(i);
+            runLen.push(j - i);
+            i = j;
+        } else {
+            i++;
+        }
+    }
+    const runCount = runStart.length;
+    if (runCount === 0) return text;
+
+    // suffixMax[a] = longest run over runs[a..end]; suffixMax[runCount] = 0. Lets the
+    // best closer for a given opener be found without re-scanning later runs.
+    const suffixMax = new Array(runCount + 1);
+    suffixMax[runCount] = 0;
+    for (let a = runCount - 1; a >= 0; a--) {
+        suffixMax[a] = Math.max(runLen[a], suffixMax[a + 1]);
+    }
+
+    const out = text.split('');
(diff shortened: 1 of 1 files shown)


chhoumann:
## Adversarial review verdict: PASS

Verified the fix with two independent adversarial passes - a 3-skeptic workflow (Claude) and an opposite-model panel (Codex Skeptic + Minimalist). Reviewers were prompted to **refute**, not validate.

**Equivalence (byte-for-byte vs the old regex)** - not refuted. Combined across passes: **>110M inputs** (exhaustive over `` {`, a, \r, \n, \t, emoji, :, combining-mark, space} `` up to length 8-9, exhaustive multi-run length grids, millions of randomised structured inputs, CRLF + surrogate pairs + lone surrogates). Zero divergence, zero length drift. One reviewer independently reconstructed the regex semantics into a second simulator and cross-checked it agrees with `old.js` on 300k inputs - ruling out a shared blind spot in the reference.

**Linearity** - not refuted. Timed 14+ adversarial shapes up to 16M chars, including patterns *designed so any region re-walk would explode* (doubling-gap, leftover-chain, giant-walk). All scale ~2.0x per input doubling (linear); the old regex is ~4x/doubling (the actual ReDoS). Structural reason: the opener index only advances, closer-search regions are disjoint across openers, and the suffix-max table makes each per-opener decision O(1).

**Behavioural non-regression (#1439 / #1447)** - not refuted. Length preserved exactly; `::` inside a code span is blanked so it is not migrated; a line with a code span before the colon is left verbatim (`nameIsClean` holds). On realistic notes, real fields migrate, fenced/blockquoted/list-nested code blocks are preserved, wikilink-comma values stay whole, frontmatter is untouched, and the 200k-backtick line migrates in ~4ms.

**Findings addressed**
- *(low)* "single linear pass" wording was imprecise - the masker indexes runs then walks them once (linear-time, not a single streaming pass). Comment tightened in b5796974.
- *(medium, rejected)* Reviewer suggested switching to the simpler sibling `/`[^`]*`/g` masker. Rejected: that silently downgrades a documented feature (multi-backtick spans, `` ``co`de`` ``), leaving span interiors visible to the field parser - a behavioural delta whose safety rests on an unproven argument. The mandate here is *do not regress* #1439/#1447 behaviour, and a provable byte-for-byte equivalence is the more defensible choice than a simpler-but-different variant.

PR #1478, opened 2026-07-06
docs(examples): add Templater folder-templates recipe

chhoumann:
Adds a community recipe to the Examples docs: reuse Templater's folder-template configuration from a QuickAdd Template choice, so the folder-to-template mapping is not maintained in two places.

The recipe (contributed by codedpalette in discussion 1105) is a small Templater user script that reads `folder_templates` from Templater's settings, plus a one-line shim template that `tp.file.include`s the resolved template. One QuickAdd Template choice then works for every folder that has a Templater folder template. The docs page makes two things explicit that the original recipe glossed over: the user script file must be named `folder_template.js` (it is invoked as `tp.user.folder_template`), and the lookup is exact-match only (no parent-folder fallback, unlike Templater's own trigger).

Changes:
- New page `docs/docs/Examples/Template_TemplaterFolderTemplates.md`, plus a row in the Examples overview table and a mention in the "Create structured notes" section
- Sidebar entry in `sidebars.js`
- All of the above mirrored into the live `version-2.17.2` snapshot (the site root serves the frozen version; `docs/docs` is `/next/`)

Verified with a full `pnpm build` in `docs/` - the page renders in both the root and `/next/` versions.

### DIFF
--- a/docs/docs/Examples/Template_TemplaterFolderTemplates.md
+++ b/docs/docs/Examples/Template_TemplaterFolderTemplates.md
@@ -1,0 +1,75 @@
+---
+title: "Template: Reuse Templater Folder Templates"
+---
+
+If you use [Templater's](https://github.com/SilentVoid13/Templater) folder
+templates, a QuickAdd Template choice normally needs its own template path -
+so you end up maintaining the folder-to-template mapping in two places.
+
+This recipe removes the duplication. A small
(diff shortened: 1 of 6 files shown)


chhoumann:
Withdrawing: keeping the docs focused on QuickAdd-native workflows rather than bridging into Templater settings. The recipe remains available in discussion 1105 for anyone who needs it.

Issue #1422, opened 2026-06-22
[FEATURE REQUEST] Template, Link Type Regardless of Link Placement


jbutz:
**Is your feature request related to a problem? Please describe.**
I maintain daily notes in obsidian, one of the things in those notes is a list of the meetings I've had that day. I embed my notes, which are stored in a separate folder. I use an h3 for the meeting title and then I reuse that value as part of the notes filename, without the `###` of course. I like the link placement option of "New line", I simply wish I could choose the link type as well so that I could get the embed option.

**Describe the solution you'd like**
Allow the Link Type option to be selected regardless of the Link Placement value

**Describe alternatives you've considered**
I would continue the workflow from before finding this plugin where I simply copy the note name back and add it to the daily note.

chhoumann:
Hi @jbutz, thanks for the clear writeup - this is now implemented and merged, and will ship in the next release.

You can now pick the **Link type** independently of the **Link placement** for the current note. In your Capture or Template choice, under "Link to file":

- **Link destination:** Current note
- **Link placement:** New line
- **Link type:** Embed

That drops `![[Your Note]]` on its own line, which should match the daily-note meeting-embed workflow you described - reuse the note title as the filename, then embed it on a new line instead of copying it over by hand.

The **Link type** option now appears for every current-note body placement (Replace selection, After selection, End of line, and New line), not just Replace selection.

A couple of notes:
- **In frontmatter property** and the **Specified note** destination stay link-only for now, since an embed inside a property value doesn't really apply.
- It works regardless of whether your vault uses wikilinks or Markdown links - the embed always comes out as a proper `![[...]]` transclusion.

Once it lands in a release, give it a try and let me know if anything behaves differently than you'd expect. Thanks again for the suggestion!

PR #1491, opened 2026-07-06
feat(append-link): keep the selected text as the link display text

### LINKED ISSUE #1479: Append link: option to use the original selection as the link display text
When a Template or Capture choice appends a link with the `replaceSelection` or `afterSelection` placement, the selected text is discarded and the inserted link has no alias. `buildFileLinkText` (src/utils/fileLinks.ts) calls `app.fileManager.generateMarkdownLink(file, sourcePath)` without the alias argument, which `generateMarkdownLink` supports.

Proposal: add a `displayText` option (`'none' | 'selection'`) to `AppendLinkOptions`, honored for the selection-based placements, so selecting "Meeting with Mark" and running the choice produces `[[20240101 Meeting with Mark|Meeting with Mark]]`.

Requested in discussion #640 (also covers the zettelkasten variant raised there).

chhoumann:
Adds a **Link display text** option to the append-link settings of Template and Capture choices. With `Selected text` chosen, the selection-based placements (**Replace selection**, **After selection**) keep the highlighted text as the inserted link's alias: select `Meeting with Mark`, run a Template choice named `20240101 {{selected}}`, and the selection becomes `[[20240101 Meeting with Mark|Meeting with Mark]]` instead of a bare `[[20240101 Meeting with Mark]]`. This is the exact workflow from discussion #640, including the zettelkasten block-extraction variant.

Closes #1479. Refs discussion #640.

## Design

- **Schema.** `AppendLinkOptions.displayText?: 'none' | 'selection'` - a string enum matching the option design language from the 2.13 appendLink rework and the #1422 embed work (`linkType`, `frontmatterHandling`), leaving room for future values (`title`, `custom`) without a schema break. Selection-capable placements are an exported capability set (`placementSupportsSelectionAlias`), the same pattern as `EMBED_CAPABLE_PLACEMENTS`, consumed by the normalizer and the UI so they cannot drift.
- **Normalization is a strict allowlist.** `'selection'` survives only for a plain link into the active note with a selection-anchored placement. Legacy booleans, malformed/unknown values (imported settings, third-party scripts), embeds, frontmatter, cursor-anchored placements, and specified-note destinations all normalize to `'none'`. The normalized type exposes a non-optional `displayText`, and `insertFileLinkToActiveView` re-normalizes internally so raw callers cannot bypass the sanitization.
- **Alias safety is QuickAdd's job.** Obsidian's `generateMarkdownLink` performs no alias sanitization (verified live against 1.13's parser via `metadataCache`). Rules, all grounded in live probes:
  - Newline runs collapse to a single space (multi-line zettelkasten selections stay usable); empty selections degrade to a plain link.
  - Markdown-link vaults escape `\`, `[`, `]` (renders display-faithfully, every alias representable).
  - Wiki vaults cannot escape (backslash is literal in wikilinks), so unrepresentable aliases - containing `]]` or `[[` (a nested `[[Other]]` hijacks the outer link's target entirely) or ending in `]` - are **dropped** rather than mutating the user's text. Single `[`, `]`, `|` pass through (safe per the live parser probes).
- **Insertion mechanics.** The selection placements derive each cursor's link from that selection's own text, applied in one atomic `editor.transaction` (single undo step). `replaceSelection` collapses each cursor after its inserted link, matching `editor.replaceSelection` - CodeMirror's default mapping would otherwise leave the link selected and the next keystroke would destroy it. All no-alias code paths are byte-identical to before (the no-alias `generateMarkdownLink` call still passes 2 args).
- **UI.** The dropdown shows only when the conditions hold; switching placement away and back resets it, intentionally consistent with the `linkType` reset. `nextOptions` falls back to the *normalized* value so malformed stored values are not written back on unrelated changes.

## Tradeoffs

- `afterSelection` + `Selected text` visually duplicates the text (`Meeting with Mark [[X|Meeting with Mark]]`) - opt-in, explicitly requested in #1479.
- Capture flows that consume the selection before link insertion (capture to the active file at the cursor) degrade gracefully to a plain link: the invariant is *the alias always equals the text the link replaces*, read at insertion time.
- The focused-property runtime flow routes to `appendLinkToFrontmatterProperty` before this code and intentionally never aliases.
- `endOfLine`/`newLine` are cursor-anchored, not selection-anchored, and stay out of scope; the capability set makes extending them a deliberate one-line decision.

## Validation

Design was adversarially reviewed before coding (3 lenses + 2 adversarial reviewers); the implementation got a second adversarial pass whose findings are fixed in the last commit (a quadratic-backtracking alias regex of the #1444/#1455/#1462 opener-flood class, a raw-caller `requireActiveFile` regression, an empty-selections edge, and a UI write-back of malformed values).

Everything below was proven **live** in an isolated e2e vault (Obsidian 1.13, `pnpm run provision:e2e-vault` + `obsidian:e2e`), after first reproducing the aliasless link on master:

| Scenario | Result |
| --- | --- |
| Exact #640 flow (select `Meeting with Mark`, template `20240101 {{selected}}`) | `[[20240101 Meeting with Mark\|Meeting with Mark]]`, cursor after link |
| Default settings (regression) | `[[20240101 Sync One]]` - unchanged |
| `afterSelection` | `Sync Two[[20240101 Sync Two\|Sync Two]]` |
| Embed + displayText requested | `![[20240101 Sync Three]]` - alias normalized away |
| Multi-line selection | `[[Zettel\|Confirmation Bias is a tendency of the mind]]` |
| Hostile selections (`a]]b`, `see [[Other]] ref`) | plain link, target intact |
| Safe specials (`array[0] a\|b`) | alias kept verbatim |
| Empty selection | plain link at cursor |
| Two cursors, different selections | per-selection aliases, cursors collapsed after each link, **one** undo restores all |
| Markdown-link vault | `[Meeting with Mark](Zettel%20Note11.md)`; `array[0]` -> `[array\[0\]](...)` |
| Capture to another note | selection -> `[[CaptureTarget\|Interesting idea]]`, capture landed |
| Capture to active file at cursor | degrades to plain link (selection consumed by the capture) |
| Settings UI | row renders with `Default`/`Selected text`, hides for `endOfLine`/embed, lossy round-trip shows `Default` |

Unit coverage: normalization matrix (legacy/malformed/unknown values), alias sanitizer vectors, per-selection insertion mechanics against a real text-model harness (cursor math, reversed selections, atomicity), linear-time regression test for the alias regex, and UI visibility/threading tests. Full suite: 3670 passing.

## Summary by CodeRabbit

* **New Features**
  * Added “Link display text” support so selected text can be reused as a plain link’s alias in compatible selection-based placements.
  * Added safer selection-derived alias handling, including Markdown-style link formatting.
* **Bug Fixes**
  * Multi-line/whitespace-heavy selections are normalized for cleaner display.
  * Embeds and unsupported placements/destinations automatically fall back to plain link behavior.
* **Documentation**
  * Updated CaptureChoice and TemplateChoice documentation with examples and dropdown behavior.
* **Tests**
  * Expanded test coverage for display-text normalization, insertion behavior, and alias generation.

### DIFF
--- a/docs/docs/Choices/CaptureChoice.md
+++ b/docs/docs/Choices/CaptureChoice.md
@@ -237,6 +237,8 @@ The Capture builder is grouped into sections: **Location**, **Position**, **Link
 
     For any **Current note** body placement — **Replace selection**, **After selection**, **End of line**, and **New line** — an extra _Link type_ dropdown appears, letting you choose **Link** (`[[Note]]`) or **Embed** (`![[Note]]`). An embed transcludes the linked note's contents at the placement position (for example **New line** + **Embed** drops `![[Note]]` on its own line; the inline placements insert it on the same line). The **In frontmatter property** placement and the **Specified note** destination stay link-only.
 
+    For the selection placements (**Replace selection**, **After selection**) with the **Link** type, a _Link display text_ dropdown chooses what the inserted link displays. **Selected text** keeps the highlighted text as the link's display text — selecting `Meeting with Mark` and capturing to `20240101 Meeting with Mark` inserts `[[20240101 Meeting with Mark|Meeting with Mark]]`. With nothing selected (or when the selection can't be represented safely inside a link), QuickAdd inserts the plain link instead. Multi-line selections are collapsed to a single line for the display text, and vaults using Markdown-style links get `[Meeting with Mark](20240101%20Meeting%20with%20Mark.md)`.
+
     For the **Specified note** destination, choose an existing Markdown file. QuickAdd appends a normal link at the bottom of that file after each successful capture. It does not create the index file, insert under a heading, update properties, or remove duplicate links.
 
 -   _Copy link to clipboard_ copies a link to the captured file after the Capture
--- a/docs/docs/Choices/TemplateChoice.md
+++ b/docs/docs/Choices/TemplateChoice.md
@@ -86,6 +86,8 @@ of the value, and list properties receive a new list item.
 
 **Link type**. Shown for any **Current note** body placement — **Replace selection**, **After selection**, **End of line**, and **New line**. Choose whether QuickAdd inserts a **Link** (`[[Note]]`) or an **Embed** (`![[Note]]`). An embed transcludes the linked note's contents at the placement position, so for example **New line** + **Embed** drops `![[Note]]` on its own line. The inline placements (**After selection**, **End of line**) insert the embed inline on the same line. The **In frontmatter property** placement and the **Specified note** destination stay link-only.
 
+**Link display text**. Shown for the selection placements (**Replace selection**, **After selection**) with the **Link** type. Chooses what the inserted link displays. **Selected text** keeps the highlighted text as the link's display text: select `Meeting with Mark`, run a Template choice whose file name format is `20240101 {{selected}}`, and the selection becomes `[[20240101 Meeting with Mark|Meeting with Mark]]` instead of a bare `[[20240101 Meeting with Mark]]`. With nothing selected (or when the selection can't be represented safely inside a link), QuickAdd inserts the plain link instead. Multi-line selections are collapsed to a single line for the display text, and vaults using Markdown-style links get `[Meeting with Mark](20240101%20Meeting%20with%20Mark.md)`.
+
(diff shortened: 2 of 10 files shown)


chhoumann:
@coderabbitai review

PR #1513, opened 2026-07-10
chore(e2e): adopt shared e2e fixtures

chhoumann:
## What

Rebuilds QuickAdd's E2E test-side plumbing on `obsidian-e2e@0.8.1`'s shared fixtures, removing the hand-rolled env resolution, vault-lock, and client wiring that each test file leaned on. QuickAdd is the last and shape-wise trickiest consumer (it never had a central harness factory), so the migration is a per-file decision rather than a blanket rewrite.

Follows the MetaEdit (#1511 runner, plus the harness adoption in chhoumann/MetaEdit#180) and PodNotes migrations.

## Per-file decision

| File | Decision | Why |
| --- | --- | --- |
| `e2eVault.ts` | Rebuilt on `resolveObsidianEnvOptions` + `verifyVaultPath` | Shared env resolution (canonical `OBSIDIAN_E2E_*` with legacy `QUICKADD_E2E_*` fallback, per-client HOME injection) and vault-path verification. Keeps its exported names so the per-file suites import unchanged. |
| `apply-template-to-active-note.test.ts` | Adopt `createPluginHarness`; **keep the poll** | Seeds no cross-test plugin data, so the suite-scoped lock + sandbox + reload with per-test data restore fits. The template-application helper keeps its window-global poll instead of `evalJsonAsync`: template application emits `QuickAdd:` notices into the eval channel that corrupt the JSON envelope (verified live - `evalJsonAsync` failed A01/A04 with `Unexpected token 'Q', "QuickAdd: "... is not valid JSON`). |
| `locked-leaf-file-opening.test.ts` | Env-resolver + migrate poll -> `evalJsonAsync` | Bespoke split-pane lifecycle (re-seeds data per test) stays. Its window-global layout poll migrates cleanly to `obsidian.dev.evalJsonAsync` - an async IIFE returning the layout result with no notice noise (passes live). |
| `conditional-branch-persistence.test.ts` | Env-resolver only | Single test, no sandbox, custom `closeAllModals` teardown, synchronous-eval GUI driving. Deviates from the suite shape; logic untouched. |
| `file-exists-behavior.test.ts` | Env-resolver only | Seeds 17 choices in a `describe` `beforeAll` and asserts them across 15+ tests. The harness rolls `data.json` back after every test, which would wipe the seed. |
| `macro-member-access.test.ts` | Env-resolver only | Same describe-level seed persisting across tests. |
| `scorecard-composed-flows.test.ts` | Env-resolver only | Same describe-level seed persisting across tests. |
| `template-property-links.test.ts` | Env-resolver only | Same describe-level seed persisting across tests. |

The six env-resolver-only suites need no source edits - they inherit the dedup through the rebuilt `e2eVault.ts` imports (`createQuickAddObsidianClient`, `acquireQuickAddVaultRunLock`).

## Line delta

`134 insertions(+), 186 deletions(-)` across the three touched test-side files (net -52); `e2eVault.ts` alone: +50 / -45.

## Gates

- `pnpm run build-with-lint` - green.
- `pnpm run test` (unit) - 3827 passed, 37 skipped.

## Live E2E vs baseline

Full suite via the shared runner (`start:e2e-obsidian` -> HOME remap -> `test:e2e`): **44/45**, matching the runner-migration baseline. The only failure is `conditional-branch-persistence.test.ts` (`cfg=false` - the settings GUI does not render the Configure button in time). Diagnosed, not waved through: it reproduces **identically on `master`** against the same built runtime (this PR makes no `src/` changes), confirming it is pre-existing and environmental, not a regression. `apply-template` (8/8) and `locked-leaf` both pass.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test stability by standardizing setup with a shared harness.
  * Refined template-application and file-opening test flows to use context-aware setup and cleaner data retrieval.
  * Updated layout-state handling in locked leaf scenarios for more reliable assertions.
  * Kept vault validation coverage while improving how test vault paths are resolved and verified.
* **Chores**
  * Upgraded the end-to-end testing toolkit dependency for improved compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "jsdom": "29",
     "moment": "2.30.1",
     "obsidian": "1.13.1",
-    "obsidian-e2e": "^0.7.0",
+    "obsidian-e2e": "^0.8.1",
     "semantic-release": "^25.0.5",
     "svelte": "^5",
     "svelte-check": "^4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -93,8 +93,8 @@ importers:
         specifier: 1.13.1
         version: 1.13.1(@codemirror/state@6.5.0)(@codemirror/view@6.38.6)
       obsidian-e2e:
-        specifier: ^0.7.0
-        version: 0.7.0(vite-plus@0.1.24(@types/node@26.1.0)(@vitest/coverage-v8@4.1.10)(esbuild@0.28.1)(jiti@2.6.1)(jsdom@29.1.1)(svelte@5.56.4(@typescript-eslint/types@8.62.1))(typescript@6.0.3)(vite@8.1.3(@types/node@26.1.0)(esbuild@0.28.1)(jiti@2.6.1)(yaml@2.9.0))(yaml@2.9.0))
+        specifier: ^0.8.1
+        version: 0.8.1(vite-plus@0.1.24(@types/node@26.1.0)(@vitest/coverage-v8@4.1.10)(esbuild@0.28.1)(jiti@2.6.1)(jsdom@29.1.1)(svelte@5.56.4(@typescript-eslint/types@8.62.1))(typescript@6.0.3)(vite@8.1.3(@types/node@26.1.0)(esbuild@0.28.1)(jiti@2.6.1)(yaml@2.9.0))(yaml@2.9.0))(vitest@4.1.10)
       semantic-release:
         specifier: ^25.0.5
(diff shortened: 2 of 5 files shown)


chhoumann:
Addressed the nitpick in 0b32b4fc: `resolveObsidianEnvOptions({ legacyPrefix: "QUICKADD" })` is now called once at module scope and the stored result feeds both the client options and the harness.

PR #1526, opened 2026-07-15
feat: add command to open QuickAdd settings page

davidjoydong:
##  Summary
Added a new command `Open QuickAdd settings` that allows users to directly open the QuickAdd plugin's settings page from the command palette or CLI.
This is especially useful when creating complex Macros, where one of the steps may require guiding the user to open the plugin settings.

## Changes
- Added new command `open-quickadd-settings` in `onload()` method of `src/main.ts`
- Command uses internal `app.setting.open()` + `openTabById()` to open the plugin settings tab

## Testing / validation
- Ran `pnpm run build` successfully
- Tested in a dedicated development vault
- Command appears in command palette and correctly opens QuickAdd settings page
- Verified no other files were affected

## Checklist

- [ ] Linked any related issue(s).
- [ ] Noted release/migration impact, if any.

## Summary by CodeRabbit

* **New Features**
  * Added a command to open QuickAdd settings directly (“Open settings”); it will try automatic opening and, if unavailable, prompt you to open **Settings → QuickAdd** manually.
* **Bug Fixes**
  * Updated the “missing template folder configuration” flow to use the same settings-opening behavior with a manual fallback.
* **Documentation**
  * Updated QuickAdd command labels so the new settings command shows correctly in the command list.
* **Tests**
  * Added automated tests for the settings-opening behavior and fallback messaging.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -112,7 +112,7 @@ export default class QuickAdd extends Plugin {
 			},
 		});
 
-		// add: 
+		// add: Allows users to quickly open QuickAdd settings from the command palette
 		this.addCommand({
 			id: "open-quickadd-settings",
 			name: "Open QuickAdd settings",
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,6 +1,6 @@
 /** biome-ignore-all assist/source/organizeImports: Import order is critical to prevent circular dependencies - ChoiceExecutor must load before dependent classes */
 import type { Debouncer } from "obsidian";
-import { Plugin, TFile, debounce } from "obsidian";
+import { Plugin, TFile, debounce, Notice } from "obsidian";
 import { QuickAddSettingsTab } from "./quickAddSettingsTab";
 import { DEFAULT_SETTINGS } from "./settings";
 import type { QuickAddSettings } from "./settings";
@@ -46,7 +46,7 @@ import {
 	parseCallbackTargets,
 	type CallbackTargets,
 } from "./uri/uriCallback";
-import { runTemplateFromFolder } from "./engine/runTemplateFromFolder";
+import { runTemplateFromFolder, tryOpenPluginSettings } from "./engine/runTemplateFromFolder";
 
 // Parameters prefixed with `value-` get used as named values for the executed choice
 type CaptureValueParameters = { [key in `value-${string}`]?: string };
@@ -109,17 +109,6 @@ export default class QuickAdd extends Plugin {
 				ChoiceSuggester.Open(this, this.settings.choices, {
 					includeTemplateFolderRow: true,
 				});
-			},
-		});
-
-		// add: Allows users to quickly open QuickAdd settings from the command palette
-		this.addCommand({
-			id: "open-quickadd-settings",
-			name: "Open QuickAdd settings",
-			callback: () => {
-				const setting = (this.app as any).setting;
-				setting?.open?.();
-				setting?.openTabById?.(this.manifest.id);
 			},
 		});
 
@@ -221,6 +210,18 @@ export default class QuickAdd extends Plugin {
 				void fn();
 			},
 		});
+
+		this.addCommand({
+            id: "openQuickAddSettings",
+            name: QUICK_ADD_COMMAND_LABELS.openSettings,
+            callback: () => {
+                if (!tryOpenPluginSettings(this.app, this.manifest.id)) {
+                    new Notice(
+                        "QuickAdd: Unable to open settings automatically. Open Settings -> QuickAdd manually."
+                    );
+                }
+            },
+        });
 
 		this.registerObsidianProtocolHandler("quickadd", async (e) => {
 			const parameters = e as unknown as UriParameters;
--- a/src/main.ts
+++ b/src/main.ts
@@ -9,6 +9,7 @@ import { ConsoleErrorLogger } from "./logger/consoleErrorLogger";
 import { GuiLogger } from "./logger/guiLogger";
 import { LogManager } from "./logger/logManager";
 import { reportError, withErrorHandling } from "./utils/errorUtils";
+import { tryOpenPluginSettings } from "./utils/openPluginSettings";
 import { StartupMacroEngine } from "./engine/StartupMacroEngine";
 import { ChoiceExecutor } from "./choiceExecutor";
 import type IChoice from "./types/choices/IChoice";
@@ -46,7 +47,7 @@ import {
 	parseCallbackTargets,
 	type CallbackTargets,
 } from "./uri/uriCallback";
-import { runTemplateFromFolder, tryOpenPluginSettings } from "./engine/runTemplateFromFolder";
+import { runTemplateFromFolder } from "./engine/runTemplateFromFolder";
 
 // Parameters prefixed with `value-` get used as named values for the executed choice
 type CaptureValueParameters = { [key in `value-${string}`]?: string };
--- a/src/main.ts
+++ b/src/main.ts
@@ -212,16 +212,16 @@ export default class QuickAdd extends Plugin {
 		});
 
 		this.addCommand({
-            id: "openQuickAddSettings",
-            name: QUICK_ADD_COMMAND_LABELS.openSettings,
-            callback: () => {
-                if (!tryOpenPluginSettings(this.app, this.manifest.id)) {
-                    new Notice(
-                        "QuickAdd: Unable to open settings automatically. Open Settings -> QuickAdd manually."
-                    );
-                }
-            },
-        });
+			id: "openQuickAddSettings",
+			name: QUICK_ADD_COMMAND_LABELS.openSettings,
+			callback: () => {
+				if (!tryOpenPluginSettings(this.app, this.manifest.id)) {
+					new Notice(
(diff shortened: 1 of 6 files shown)


chhoumann:
Hey @DavidJoyDong, thank you for taking the time to contribute this. I tested the branch locally and confirmed that the command works as described.

Before merging it, I want to step back and better understand the product use case. On macOS, `Cmd+,` already opens Obsidian’s settings and returns to the previously selected settings tab. For someone who was recently configuring QuickAdd, that provides almost the same shortcut.

Could you share the specific workflow or problem that motivated this command? For example, is the goal a dedicated hotkey, use on mobile, or invoking it from a macro? Understanding the concrete situation would help me evaluate whether this deserves a permanent place in QuickAdd’s command palette.

I am also somewhat cautious because opening a specific settings tab relies on Obsidian’s undocumented `app.setting` API. QuickAdd already uses that internally when it can contextually guide someone to missing configuration, but I would prefer not to expand that dependency without a clear user benefit.

My current inclination is not to add a general-purpose command unless there is a stronger use case that the existing settings shortcut does not cover. This is not a criticism of the implementation, which works, but a question of keeping QuickAdd’s command surface focused and using Obsidian internals only where they solve a concrete problem.

Thanks again for the contribution. I would be happy to reconsider based on the workflow you had in mind.

davidjoydong:
### I agree with you
Thank you for the reply! I’m a new QuickAdd user and I really like the plugin — I think it has a lot of potential.

> I am also somewhat cautious because opening a specific settings tab relies on Obsidian’s undocumented `app.setting` API. QuickAdd already uses that internally when it can contextually guide someone to missing configuration, but I would prefer not to expand that dependency without a clear user benefit.
> 
> My current inclination is not to add a general-purpose command unless there is a stronger use case that the existing settings shortcut does not cover. This is not a criticism of the implementation, which works, but a question of keeping QuickAdd’s command surface focused and using Obsidian internals only where they solve a concrete problem.

I hadn’t fully considered these points before submitting the PR. QuickAdd is somewhat special because it lets users register their own commands, and people’s naming habits vary widely. Adding this command could indeed end up mixed in among user-defined ones, which might feel like noise to long-time users.

I also understand your caution around the undocumented `app.setting` API.

I personally wouldn’t mind — and would actually like — having the command, but **I agree with you** that we should be careful about adding general-purpose commands and take all users into account.

One **alternative** I thought of is to provide a preference setting that lets users choose whether to register this command (with it disabled by default).

**Another approach** would be to implement this functionality inside a QuickAdd macro (similar to how the “Wait for” button works), so that users can register and use it themselves if they want it.

### Why I submitted this feature

#### Fluid, distraction-free configuration

In Obsidian, some plugins only need to be configured once. Others need to be adjusted repeatedly as you use them, so they benefit from a quicker settings entry point.

> Before merging it, I want to step back and better understand the product use case. On macOS, `Cmd+,` already opens Obsidian’s settings and returns to the previously selected settings tab. For someone who was recently configuring QuickAdd, that provides almost the same shortcut.

`Cmd+,` is the default way to open settings when a plugin doesn’t provide its own command, but I find it inconvenient and it breaks the flow of working in Obsidian (you have to set aside whatever note you’re working on to focus on configuration). Specifically, the sequence is: `Cmd+,` → type `quickadd`(I have many plugins installed and prefer using search rather than scrolling and clicking through the list, as it feels smoother) → `↓↓↓` → `↵`.

One of my favorite commands is **“Style Settings: Show style settings view”**. It lets me tweak the interface while I’m still in the middle of using Obsidian — smoothly, lightly, and with almost no interruption. Tools like QuickAdd that support macros can make this even more fluid. I’ve even set up a macro for it:

```
Style Settings: Show style settings view
wait for 300 ms
Move current tab to new window
```
20260722_101957_Pixpin

Other plugins have similar dedicated commands:
- “Style Settings: Show style settings view”
- “Note Toolbar: Plugin settings”
- “Commander: Open Commander Settings” **(Commander can also register commands, yet it chose to provide only this one general-purpose command)**
- “Keyboard Analyzer: Open keyboard shortcuts view (new tab)”

#### Most concrete use case
> Could you share the specific workflow or problem that motivated this command? For example, is the goal a dedicated hotkey, use on mobile, or invoking it from a macro? Understanding the concrete situation would help me evaluate whether this deserves a permanent place in QuickAdd’s command palette.

I built a “template for templates.” I want a macro that:
1. Creates a new template from that meta-template, and
2. Immediately opens QuickAdd settings so I can configure the trigger for the new template.

That may sound a bit abstract, so here’s a concrete example from my personal life vault. I use it to track my various hobbies, many of which generate large numbers of similar entries.

- I first created a “entry-note template template.”
- When I get into cooking, I use that meta-template to create a “Recipe template,” then open QuickAdd to set up its trigger.
- When I get into keyboards, I create a “Keyboard template” the same way and configure the trigger.
- When I get into football players, I create a “Football player template” (which needs extra domain-specific properties beyond a normal person entry) and configure the trigger.
- When I get into bags, I create a “Bag template” and configure the trigger.
- …and so on.

This pattern has become regular enough that I’d like to accelerate it with a macro, and “open QuickAdd settings” is one of the steps.

### Additional thoughts

Once Obsidian ships its CLI, I believe the richer a plugin’s command surface is, the higher its ceiling becomes — anything registered as a command can be triggered via the CLI, which opens up many more possibilities. QuickAdd’s ability to let users register their own commands is already a powerful foundation for that.

src/main.ts
```diff
@@ -112,6 +112,17 @@ export default class QuickAdd extends Plugin {
 			},
 		});
 
+		// add: Allows users to quickly open QuickAdd settings from the command palette
```
chhoumann: This comment repeats what the command definition already communicates. Let's remove it.


```suggestion
```

src/main.ts
```diff
@@ -112,6 +112,17 @@ export default class QuickAdd extends Plugin {
...
+		this.addCommand({
+			id: "open-quickadd-settings",
+			name: "Open QuickAdd settings",
```
chhoumann: QuickAdd’s built-in command IDs use camelCase, and Obsidian automatically prefixes command names with the plugin name. Please also add openSettings: "Open settings" to `QUICK_ADD_COMMAND_LABELS`.


```suggestion
			id: "openQuickAddSettings",
			name: QUICK_ADD_COMMAND_LABELS.openSettings,
```

src/main.ts
```diff
@@ -112,6 +112,17 @@ export default class QuickAdd extends Plugin {
...
+			callback: () => {
+				const setting = (this.app as any).setting;
+				setting?.open?.();
+				setting?.openTabById?.(this.manifest.id);
+			},
```
chhoumann: Optional chaining here would turn an Obsidian API change into a silent no-op for the user.
Could you move the existing settings-opening logic from `runTemplateFromFolder.ts` into a shared typed helper, such as t`ryOpenPluginSettings(app, pluginId): boolean`? It should return false and log the error when the internal API is unavailable or throws.
The command can then provide useful feedback:

```suggestion
			callback: () => {
				if (!tryOpenPluginSettings(this.app, this.manifest.id)) {
					new Notice(
						"QuickAdd: Unable to open settings automatically. Open Settings → QuickAdd manually.",
					);
				}
			},
```

This will also require importing `Notice` and the shared helper. The existing `runTemplateFromFolder` caller can ignore the result because it already shows instructions to the user before attempting navigation.

chhoumann (changes requested):
Thanks for the thoughtful explanation. The concrete use case makes sense to me, and I’m happy to move forward with the feature. I’ve left a few small inline suggestions around naming and safely containing the undocumented Obsidian API.

davidjoydong:
Hi @chhoumann, thank you so much for the detailed and helpful feedback. I've updated the PR based on all your suggestions.

Here is a summary of the changes:
- Extracted settings-opening logic into a shared, robust helper `tryOpenPluginSettings`.
- Wrapped the internal API call in a try-catch block to prevent silent failures.
- Added a `Notice` feedback in `main.ts` for users if the internal Obsidian API is unavailable.
- Removed redundant command comments and updated command ID to camelCase.
- Centralized command label using `QUICK_ADD_COMMAND_LABELS`.

chhoumann:
Hi @DavidJoyDong,

Thanks, this is very close, and the updated command works correctly in my local Obsidian test.

Could you move this generic helper into a shared module, such as `src/utils/openPluginSettings.ts`, and route the two errors through QuickAdd’s logger instead of `console.error`? Keeping it in `runTemplateFromFolder.ts` means `main.ts` depends on a template-specific engine module for unrelated settings navigation.

There is also some trailing whitespace on the blank line above the availability check.

chhoumann:
One small CI issue: `src/main.commandLabels.test.ts` asserts the complete `QUICK_ADD_COMMAND_LABELS` object, so it now needs the new label added to its expected value:

```ts
openSettings: "Open settings",
```

davidjoydong:
Hi, I've updated the code as requested. I also followed coderabbitai's follow-up suggestions and made the further corrections. Please let me know if anything else is needed. Thanks!

chhoumann (approved):
Thanks for working through the feedback so thoughtfully. Approved! Great job on this.

Issue #1544, opened 2026-07-25
[FEATURE REQUEST] Capture builder describes one field three ways ("Capture to" / "File path / format" / "File name format")


chhoumann:
**Is your feature request related to a problem? Please describe.**

The top of the Capture builder uses four different labels for what is effectively one decision, and the one field you actually type into is the least well labelled of them.

Reading top to bottom on a fresh Capture choice:

| Row | Kind | Text |
| --- | --- | --- |
| 1 | heading-ish setting, **no control** | **Capture to** - "Vault-relative path, #tag, or property:field=value. Supports format syntax (use trailing '/' for folders)." |
| 2 | toggle | **Capture to active file** |
| 3 | setting, **no control** | **File path / format** - "Choose a file, folder, #tag, property:field=value, or format syntax (e.g., {{DATE}})" |
| 4 | label | **Preview:** |
| 5 | text input | placeholder: **"File name format"** |

So "Capture to" and "File path / format" are two descriptions of the same thing, neither is attached to a control, and the input that actually holds the capture target advertises itself as a *file name* format when it wants a *path*.

I typed `Daily/{{DATE}}.md` into it without being confident it was the right box. The live preview resolving to `Daily/2026-07-25.md` is what told me I had guessed right, not the labels.

**Describe the solution you'd like**

Collapse rows 1, 3 and 5 into a single labelled field:

```
Capture to
Vault-relative path, #tag, or property:field=value. Supports format syntax
(use trailing '/' for folders).
[ e.g. Daily/{{DATE}}.md                              ]
Preview: Daily/2026-07-25.md
```

Specifically:

- One label, one description, one input.
- Change the input placeholder from "File name format" to something path-shaped, e.g. `Daily/{{DATE}}.md`, which doubles as a worked example.
- Keep "Capture to active file" as the toggle that disables/hides the path field.

**Describe alternatives you've considered**

Leaving the structure and only fixing the placeholder. That helps, but the duplicated "Capture to" / "File path / format" descriptions would still read as two separate settings.

**Additional context**

Found during a new-user walkthrough on a clean vault, QuickAdd 2.19.1 / Obsidian 1.13.0. Related to the "Preview:" placement issue, which affects the same block.

chhoumann:
Filed as part of one first-run walkthrough of QuickAdd 2.19.1 (master @ af70dcb6) on Obsidian 1.13.0, in a clean vault with zero choices and no other community plugins. The full set:

**Bugs**
[...]

Issue #1541, opened 2026-07-25
[FEATURE REQUEST] No link to the documentation anywhere in the plugin UI


chhoumann:
**Is your feature request related to a problem? Please describe.**

There is no link to the documentation anywhere in QuickAdd's UI. A search of the whole source finds exactly one docs URL, buried in the "no scripts found" notice inside the Macro builder:

https://github.com/chhoumann/quickadd/blob/master/src/gui/MacroGUIs/noScriptsFoundNotice.ts#L35

The settings page even contains the text *"See One-page Inputs in the docs."* as plain, unlinked prose.

For a plugin whose entire power surface is syntax that must be learned (`{{VALUE}}`, `{{DATE}}`, `{{VDATE}}`, `|type:`, format syntax, capture targets, one-page inputs), an unreachable manual is the single biggest onboarding gap.

**Describe the solution you'd like**

Add docs entry points at the places where a user is actually stuck:

1. A "Documentation" link in the QuickAdd settings header (or next to the "Choices & Packages" heading).
2. A link in the "No choices yet" empty state.
3. Contextual links from the builders: the Capture format field, the file-path/format field, and the one-page input override setting.
4. Make the existing "See One-page Inputs in the docs" string an actual link.

**Describe alternatives you've considered**

Relying on the README and the community. That works for people who go looking, but not for someone who opened the settings tab and is staring at a format field with no idea what tokens exist.

**Additional context**

Found during a new-user walkthrough. Related: the token autocomplete discoverability issue, which is the same problem from the other direction (the in-app help that does exist is hidden).

chhoumann:
Filed as part of one first-run walkthrough of QuickAdd 2.19.1 (master @ af70dcb6) on Obsidian 1.13.0, in a clean vault with zero choices and no other community plugins. The full set:

**Bugs**
- #1538 - Capture at "Top of file" removes the blank line after frontmatter
- #1539 - "New folder" opens a rename modal titled "Choice name"

**Getting from "installed" to "first working choice"**
- #1540 - First run: "QuickAdd: Run" with no choices shows a bare "No results found"
- #1541 - No link to the documentation anywhere in the plugin UI
- #1542 - Format token autocomplete is undiscoverable, and its entries lack descriptions
- #1547 - "Export package…" is offered on a fresh install with nothing to export

**Builder clarity**
- #1543 - "Preview:" renders above its field and shows an empty label when there is nothing to preview (see also #1158)
[...]

PR #1561, opened 2026-07-26
feat(builder): advertise the format-token autocomplete

### LINKED ISSUE #1542: [FEATURE REQUEST] Format token autocomplete is undiscoverable, and its entries lack descriptions

Typing `{{` in a format field pops up a token autocomplete. It is a genuinely great feature, and it is effectively invisible: nothing in the UI hints that it exists. The Capture format textarea's placeholder is just "Format", and the field description is "Set the format of the capture."

I only found it because I typed `{{` on a hunch while looking for the token syntax. A new user with no prior QuickAdd exposure has no reason to try that, and no other in-app way to learn which tokens exist.

Once it does open, the list itself is hard to act on:

```
{{DATE:}}   {{date}}   {{time}}   {{name}}   {{value}}
{{mvalue}}  {{selected}}  {{clipboard}}  {{RANDOM:}}
```

- No descriptions. What is `{{mvalue}}`? What does `{{selected}}` capture, and how does it differ from `{{value}}`?
- Casing is inconsistent: `{{DATE:}}` and `{{RANDOM:}}` are upper-case with a trailing colon, the rest are lower-case without. A newcomer cannot tell whether that is meaningful or cosmetic, or how `{{DATE:}}` differs from `{{date}}`.

1. Advertise the affordance. A hint under the format field ("Type `{{` to insert a token") or a small "Insert token" button that opens the same list.
2. Give each entry a one-line description in the suggestion row, the way the "New choice" menu already does for choice types. That menu is the clearest explanation of QuickAdd's model anywhere in the product; the same treatment here would work well.
3. Make the casing/colon convention either consistent or explained (for example, group the "takes an argument" tokens and label them as such).

Documenting it only on the docs site. That does not help unless the docs are reachable from the UI, which today they largely are not.

**Additional context**

Found during a new-user walkthrough on a clean vault. The autocomplete appears above the textarea and covers the Task / Capture format toggles while open, which is fine, but it also means it is easy to miss if you are looking below the cursor.

chhoumann:
Closes #1542. **Stacked on #1560** — review that first; this branch adds one commit.

## The problem

Typing `{{` in a format field opens a list of every token, each with a one-line
description (that part shipped in #1549). Nothing in the UI said so. The reporter
found it by typing `{{` on a hunch while looking for the syntax; a new user has
no reason to try that, and no other in-app way to learn which tokens exist.

Ask 1 was deliberately left out of #1549 because its two natural homes belonged
to siblings from the same walkthrough: #1541 (docs links in the UI) and #1544 (a
worked-example placeholder). Both have now shipped, and #1541 explicitly left
"docs links inside the builders" to this issue. So this closes #1542.

## The change

The affordance is two characters, so the honest fix is to name the two
characters, right below the caret:

> Type `{{` to insert a token · [Format syntax](https://quickadd.obsidian.guide/docs/FormatSyntax/)

"Format syntax" is the docs entry point #1541 deferred here: the token language
is the one part of the builder you cannot work out by looking at it.

| Capture to (empty field) | Capture format |
|---|---|
| hint | hint |

## Decisions worth arguing with

**It hides once the value contains `{{`.** The hint has done its job, and a line
that never goes away is exactly the permanent chrome an "Insert token" button
was rejected for in the issue thread. Stateless — no setting, no migration.

**Three of the five token fields, not all five.** Capture format, File name
format, and Capture to — the last being the only format field a brand-new
Capture builder renders, since the other two sit behind default-off toggles.

Deliberately **not** on insert-after / insert-before: their own descriptions tell
you to type a literal heading like `## Tasks`, so a token-less value is the
documented normal state there and the hint would nag forever. Not on the template
path field either — it has no token autocomplete at all.

**Not a persisted "you've seen this" flag.** It was the obvious alternative and
it is worse in four ways: `settingsStore` mutations rewrite the whole
`data.json`, so the suggester would write to disk on a keystroke; no Svelte
component subscribes to the store, so the flag would not be reactive and the
hint would linger until the builder was reopened; settings are per-vault, so
"user lifetime" is false; and one stray `{{` would delete the affordance
vault-wide with no way to bring it back.

**Not the placeholder alone.** It vanishes on the first keystroke and on every
existing choice, and cannot carry the docs link. #1544 already shipped that
pattern on "Capture to" without closing this ask.

**Not `{` as the trigger, and not open-on-focus.** Both were rejected in the
issue thread: `{` + Enter would insert a token instead of a newline in a capture
format holding JSON or code, and open-on-focus fires on every builder open and
races the file suggester on "Capture to".

## Validation

- `pnpm run build`, `pnpm lint`, full suite: **4042 passed**, 0 failed.
- 4 new component tests (copy, docs link + `rel`/`target`, hide-on-`{{`,
  show-again-when-tokens-removed).
- Verified in this worktree's isolated Obsidian 1.13.0 vault: both hints render,
  both link to the live docs URL, and the hint disappears the moment a token is
  typed.

### DIFF
--- a/src/docs.ts
+++ b/src/docs.ts
@@ -17,6 +17,7 @@ export const DOCS_BASE_URL = "https://quickadd.obsidian.guide";
 
 export const DOCS_URLS = {
 	gettingStarted: `${DOCS_BASE_URL}/docs/`,
+	formatSyntax: `${DOCS_BASE_URL}/docs/FormatSyntax/`,
 	onePageInputs: `${DOCS_BASE_URL}/docs/Advanced/onePageInputs/`,
 	userScripts: `${DOCS_BASE_URL}/docs/Choices/MacroChoice/#user-scripts`,
 	packages: `${DOCS_BASE_URL}/docs/Choices/Packages/`,
--- a/src/gui/ChoiceBuilder/CaptureChoiceForm.svelte
+++ b/src/gui/ChoiceBuilder/CaptureChoiceForm.svelte
@@ -12,6 +12,7 @@ import ChoiceNameHeader from "./components/ChoiceNameHeader.svelte";
 import ValidatedInput from "./components/ValidatedInput.svelte";
 import LabeledField from "./components/LabeledField.svelte";
 import FormatPreviewField from "./components/FormatPreviewField.svelte";
+import FormatTokenHint from "./components/FormatTokenHint.svelte";
 import AppendLinkSetting from "./components/AppendLinkSetting.svelte";
 import OpenFileSetting from "./components/OpenFileSetting.svelte";
 import FileOpeningSetting from "./components/FileOpeningSetting.svelte";
@@ -170,6 +171,7 @@ function onTemplaterAfterCaptureChange(value: boolean) {
 			requiredMessage="Capture format is required when enabled"
 			makeSuggesters={formatSuggesters}
 		/>
+		<FormatTokenHint value={choice.format.format} />
 		<FormatPreviewField value={choice.format.format} {app} {plugin} />
 	{/snippet}
 </LabeledField>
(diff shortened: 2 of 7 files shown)


chhoumann:
Superseded by #1570.

GitHub closed this permanently when #1560 merged and its base branch was deleted — a closed PR's base cannot be changed and it cannot be reopened, so the rebased work moved to a fresh PR against `master`.

Same single commit, cherry-picked onto `5e18213e`. The only file overlapping the merged preview work was `styles.css`, where the two blocks turned out to be adjacent rather than conflicting; the diff against master is byte-identical to what this PR had against its old base (7 files, +107). Re-verified live on merged master before opening #1570.

Issue #1546, opened 2026-07-25
[FEATURE REQUEST] Run-time input prompts give no indication of what is being asked or where it goes


chhoumann:
**Is your feature request related to a problem? Please describe.**

When a choice runs, the input prompt is titled with the choice's name and contains a single empty box. Nothing indicates what is being asked for or what will happen with the answer.

Running my Capture:

```
New capture
[                                        ]
                          Cancel     Ok
```

Running my Template:

```
New template
[                                        ]
                          Cancel     Ok
```

They are indistinguishable, but they want completely different things: the Template prompt wants the new note's title, and the Capture prompt wants the text to capture (`{{VALUE}}`). A new user has to run each one and inspect the result to learn which is which.

There is also no indication of the destination. For the Capture I would have loved to see that it was about to write to `Daily/2026-07-25.md` - the builder already computes and displays exactly that string as a preview, so the information exists.

**Describe the solution you'd like**

Give the run-time prompt some context:

1. A placeholder that names what is wanted, varying by prompt kind, e.g. "Note title" for a Template's title prompt and "Text to capture" for a Capture's `{{VALUE}}`.
2. Optionally a subtitle showing the resolved destination, e.g. "Capturing to Daily/2026-07-25.md" or "Creating Books/Dune.md".

Both should be derivable from the choice being executed without any new configuration.

**Describe alternatives you've considered**

- A per-choice custom prompt label. More flexible, but more configuration for the common case; a sensible default per prompt kind covers most of the value.
- Relying on the choice name. That only works if the user named the choice descriptively, and the default names ("New capture", "New template") do not.

**Additional context**

Found during a new-user walkthrough on a clean vault, QuickAdd 2.19.1 / Obsidian 1.13.0. Note the choice-picker input has the same blank-input problem before you even reach this point.

chhoumann:
Filed as part of one first-run walkthrough of QuickAdd 2.19.1 (master @ af70dcb6) on Obsidian 1.13.0, in a clean vault with zero choices and no other community plugins. The full set:

**Bugs**
- #1538 - Capture at "Top of file" removes the blank line after frontmatter
- #1539 - "New folder" opens a rename modal titled "Choice name"

**Getting from "installed" to "first working choice"**
- #1540 - First run: "QuickAdd: Run" with no choices shows a bare "No results found"
- #1541 - No link to the documentation anywhere in the plugin UI
- #1542 - Format token autocomplete is undiscoverable, and its entries lack descriptions
[...]

PR #1634, opened 2026-07-30
fix: five do-or-close cleanups - truthful Multi completion, AI prompt context, new-choice race, raw NUL bytes, honest AI docs

### LINKED ISSUE #1619: [BUG] A literal NUL byte in packageImportService.ts makes ripgrep treat the file as binary
**Describe the bug**

`src/services/packageImportService.ts` line 319 contains a **literal NUL byte** (a raw 0x00 written where `"\x00undefined"` appears, instead of an escape sequence).

Because of that byte, ripgrep classifies the entire 1174-line production service as a **binary file** and skips it by default:

```
$ rg --stats -n "applyOverridesToCommands" src/
0 matches
765 files searched
```

- for a function that exists in that file.

**Why this matters**

- Every default recursive grep over `src/` silently misses this file - including the searches coding agents and contributors run constantly. During the #1571/#1572 triage this invalidated two "certain" facts before the cause was found.
- `git grep` and `git diff` are unaffected, so the blind spot never shows up in review - which is exactly why it has survived.

**Expected**

The NUL should be written as an escape sequence (`"\x00undefined"` or equivalent) so the file is plain text again. Worth also considering a tiny lint/CI check asserting no source file contains raw control bytes, so this class cannot come back.

Found during the #1571/#1572 triage investigation.

### LINKED ISSUE #1620: [BUG] AIAssistant docs describe a default-system-prompt/model fallback that does not exist for macro commands
**Describe the bug**

`docs/src/content/docs/docs/AIAssistant.md:50` documents that a macro AI Assistant command without its own system prompt falls back to the **live** default system prompt. It does not: `AIAssistantCommand`'s constructor **snapshots** `defaults.defaultSystemPrompt` at creation time (`AIAssistantCommand.ts:34`), and `MacroChoiceEngine` passes `command.systemPrompt` with no fallback. The same applies to the default model at `:48`.

**Consequence**

Editing "Default system prompt" (or the default model) in settings silently does nothing to existing macro AI commands - only newly created commands and the `quickAddApi.ai.*` script APIs read the live default. A user following the docs will edit the default and believe all their macros changed.

**Fix shape**

Either document the snapshot behavior honestly (and when each surface reads the live default), or - if the snapshot is judged a bug - change the command to fall back at run time, which is a behavior change that needs its own decision. The docs fix is the safe first step. Anchors are frozen; keep existing `{#id}`s.

Found during the #1571/#1572 triage investigation (see #1572 for the full call-site map).

### LINKED ISSUE #1621: [BUG] AI prompt templates prompt without run context: MacroChoiceEngine skips setPromptRunContext
**Describe the bug**

`MacroChoiceEngine.ts:651-655` builds the AI command's formatter **without** calling `setPromptRunContext`. So a `{{VALUE}}` / `{{VALUE:name}}` inside an AI prompt template prompts with the **generic** modal title and a **shared** draft scope, instead of the run-time context (#1546, PR #1555) every other prompt surface now gets.

**Repro**

Macro -> AI Assistant command -> prompt template containing `{{VALUE:topic}}` -> run the macro: the prompt modal appears without the "what is being asked / where it goes" context line, and its draft key collides with the generic scope.

**Expected**

The AI path constructs its formatter with the same `PromptScope` context as Template/Capture runs, so prompts inside AI templates are labelled and draft-isolated identically.

Found during the #1571/#1572 triage investigation - a live #1546 gap.

### LINKED ISSUE #1622: [BUG] js-quickadd fences inside AI prompt templates execute on every AI call - needs a deliberate look
**Describe the behavior (deliberate-look request, possibly working as intended)**

A ` ```js quickadd ` fence inside an **AI prompt template note** executes on every AI call. The template goes through `CompleteFormatter.formatFileContent`, whose first pass is `replaceInlineJavascriptInString` - so any inline-script fence in the template runs as arbitrary JS each time the AI command fires.

This is live behavior today, not hypothetical.

**Why it deserves a deliberate look**

- Prompt templates read as "text sent to the model"; users (and package reviewers) may not expect them to be a code-execution surface.
- Prompt templates can arrive via **package import**. The package preview flags scripts and commands, but a JS fence inside a referenced template note is subtler than either.
- If it IS intended (parity with Template choices), it should be documented explicitly on the AI Assistant docs page and considered in the package-preview trust surface. If it is not intended, the AI path should run a reduced pass list (the `FormatDisplayFormatter` idiom - see the #1572 analysis for the safe-pass sketch).

Found during the #1571/#1572 triage investigation.

### LINKED ISSUE #1625: [BUG] A new choice is discarded if any settings write lands while its builder is open
Found while fixing #1608, deliberately not grown into that PR: it is on `master` today, independent of the id repair, and it is data loss on the happy path rather than the malformed one.

**The shape**

`ChoiceView.addChoiceToList` adds the new choice to the LOCAL tree and then hands off to the builder without saving first (`ChoiceView.svelte`, the `else if (!skipConfigure)` arm):

```js
choices = addChoiceToTree(choices, newChoice, targetFolderId);
...
// Doers hand off to their builder, which both names and configures the
// choice and persists the result - no eager save (avoids a double write).
await handleConfigureChoice(newChoice);
```

Meanwhile the view subscribes to `settingsStore` unfiltered and re-seeds the whole tree on every emission:

```js
settingsStore.subscribe((settings) => {
    choices = seedChoices(settings.choices);   // was: choices = settings.choices
    ...
});
```

The store does not know about `newChoice` yet - nothing saved it. So any emission that lands while the builder is open replaces the local tree with one that does not contain it, and the builder's own save then persists a tree without the choice the user just spent a minute configuring.

**Why it is reachable without the user doing anything**

`autoSyncEnabledProviders` (`main.ts`, inside `onLayoutReady`) calls `settingsStore.setState` from `modelSyncService.ts:109` a few seconds after launch. Several migrations do too. Open the settings tab quickly after starting Obsidian, click "New choice", and the window is real.

The same window applies to `handleRenameChoice` and to anything else that mutates the local tree and awaits before saving.

**Suggested shape**

Two options:

1. Save before handing off. It costs the double write the comment is avoiding, but the comment is trading a write for a data-loss window.
2. Make the subscription non-clobbering: ignore an emission whose `choices` is reference-identical to what this view last SENT, and otherwise reconcile rather than replace. The seam added in #1608 already memoizes on the raw reference, so the hook is there.

(1) is smaller and obviously correct; (2) is the real fix if there are other await-then-save paths.

**Related**

- #1608 - the seam this sits on, and where the reachability evidence came from
- #1566 - the accessor split the view already uses

### LINKED ISSUE #1630: [BUG] A Multi choice run from inside a macro is fire-and-forget, so the run reports done while its picker is still open
**Describe the bug**

A Multi choice reached from inside a Macro is fire-and-forget, so an interactive
or URI run reports its terminal result while the choice picker is still open on
the desktop - and a child choice can then run out of band, after the caller has
been told the run finished.

`ChoiceExecutor.onChooseMultiType` is not awaited and is not async
(`src/choiceExecutor.ts:143`):

```ts
this.onChooseMultiType(multiChoice);   // no await
```

and it ends at `ChoiceSuggester.Open(...)` (`src/choiceExecutor.ts:346`), which is
`new ChoiceSuggester(...).open()` (`src/gui/suggesters/choiceSuggester.ts:216`) and
returns immediately.

`quickadd:interactive` refuses a Multi as the TOP-LEVEL choice
(`registerQuickAddCliHandlers.ts`, "Multi choices cannot be run interactively via
CLI"), but that guard does not cover a Multi reached from inside a run. Three
paths do:

- `MacroChoiceEngine` `executeChoice` (a macro "Choice" command pointing at a Multi)
- `MacroChoiceEngine` `executeNestedChoice`
- `quickAddApi.executeChoice(name)`

**Reproduction sketch** (not yet run end to end - filed from a read of the code
while fixing #1614)

`quickadd:interactive choice="MyMacro"` where MyMacro's first command runs a Multi
folder. Expect the bridge to emit `{"kind":"done","ok":true}` while an unanswered
folder picker sits in the Obsidian window; answering it later executes a child
choice the caller has already stopped listening for.

**Why it is filed rather than folded into #1614**

#1614's fix routes prompts an ENGINE opens through the interactive prompt
provider. This one is not a routing gap: `ChoiceSuggester` is a GUI surface with
its own lifecycle, and making it awaitable changes semantics for the in-app path
too (today opening a Multi from a macro deliberately does not block the macro).
That is a design decision worth taking on its own, not a line inside a routing
change.

**Shape of a fix**

Two independent halves:

1. Make `onChooseMultiType` awaitable - `ChoiceSuggester.Open` resolving when the
   picked child choice completes - so a run cannot report `done` while its own
   picker is open. This is the real fix and it touches in-app behaviour.
2. Until then, refuse the case that is provably wrong for a remote caller: when
   `choiceExecutor.promptProvider` is set (or `interactive === false`), throw a
   `ChoiceAbortError` naming the Multi, instead of opening a picker nobody can
   answer.

Found while fixing #1614.

chhoumann:
The final cleanup round from the do-or-close triage: four code fixes and two documentation resolutions, one commit per fix. Smallest correct change per issue, no new abstractions.

## What changed

**Raw NUL bytes made ripgrep skip whole source files.** `packageImportService.ts` and (found during this work) `modelSyncService.ts` each carried a literal `0x00` inside a string literal, so every default recursive `rg` over `src/` silently skipped them - `git grep`/`git diff` were unaffected, which is why it survived review. Both bytes are now written as `\u0000` escape sequences (byte-identical runtime strings), and a colocated guard test walks `src/` and fails on any raw control byte other than tab/LF/CR. Closes #1619

**`{{VALUE}}` prompts inside AI prompt templates had no run context.** `MacroChoiceEngine.executeAIAssistant` built its formatter without `setPromptRunContext` - the one surface #1546 missed. It now hands over the choice name and a per-command draft scope, exactly mirroring `executeOpenFile`. Closes #1621

**A new choice was discarded if any settings write landed while its builder was open.** ChoiceView re-seeds its tree from the settings store on every store write (model auto-sync fires seconds after launch), and the doer-add arm deliberately skipped the save before opening the builder - so the unsaved choice was wiped and the builder's save-by-id persisted a tree without it. The choice is now saved before the builder opens, matching the folder and Alt-click arms. The regression test fails on the pre-fix code and passes on the fix. Closes #1625

**A Multi choice run from inside a macro was fire-and-forget.** `execute()` resolved while the picker was still open: a macro ran its remaining steps before (or instead of) the picked child, `api.executeChoice` resolved early, and a CLI run emitted `ok:true` with an unanswered picker on screen. `execute()` now awaits a completion the picker settles when the picked leaf actually finishes (threaded through nested levels and Back; only the terminal action settles it). Dismissing the picker cancels the run like any other prompt; a leaf failure still names the leaf and propagates the same error instance, so report-once keeps it to a single notice. Remote/non-interactive runs are refused up front with the standard aborted frame instead of opening a picker nobody can answer. The top-level launcher passes no completion and stays fire-and-forget. Closes #1630

**AI docs claimed a live default-model/system-prompt fallback that does not exist.** `AIAssistantCommand` snapshots both at creation; editing a default never changed existing commands. The page now says the defaults are starting values copied into new commands, lists which settings are read live, and notes that the script APIs are the surfaces that do read the live default system prompt. All `{#id}` anchors kept. Closes #1620

**`js quickadd` fences in AI prompt templates execute on every AI call - documented as deliberate.** Same trust model as Template choices, now stated as a caution on the AI Assistant page, cross-referenced from the inline-scripts intro, and spelled out in the package-import trust section (a bundled note with a runnable JS fence gets a critical can-be-run-as-code row and counts toward the review-and-acknowledge gate - verified against `packagePreview.ts`). Closes #1622

## Live verification (isolated e2e vault)

Macro `QA Outer` = [run folder `QA Folder`, then a marker step]. Mid-picker the run is honestly still in flight - picker open, no files written, promise pending:

macro waits at the picker

Picking `QA Child` ran the child first (`child-ran` at t=...122), then the outer macro's next step (`after-multi` at t=...125), then reported done. Escape instead: run rejects with "Input cancelled by user" and the marker step never runs. Non-interactive CLI now gets the truth instead of `ok:true`:

```json
{"ok":false,"command":"quickadd:run","error":"\"QA Folder\" is a folder of choices, and picking from it needs someone at the window. Run one of its choices directly, or re-run with the ui flag.","aborted":true,"choice":{"id":"qa-outer","name":"QA Outer","type":"Macro"},"durationMs":2}
```

`{{VALUE:topic}}` inside an AI prompt template now prompts with the run context line (was: bare generic prompt):

AI prompt with context line

The #1625 race, driven live through the real settings tab: choice added via the New choice menu, a real settings-store write landed while the builder was open, builder closed - the choice is in memory AND on disk (`writeLanded: true`). A prior attempt that crashed mid-flow also left its choice persisted, which is the eager save doing its job:

new choice persisted

And #1619, before/after: `rg -n applyOverridesToCommands src/` went from 0 matches (file classified binary) to finding the function.

## Review notes and accepted trade-offs

- Adversarially reviewed twice (design panel before coding, three-lens diff review before this PR); no critical or important findings survived verification.
- A macro now *waits* at a folder picker and Escape *cancels* the run - deliberate semantic changes; they make Multi behave like every other prompt surface mid-run.
- While a picker is open the executor holds its execution scope, so a concurrent run on the same long-lived API executor inherits the first run's trigger context. This is the pre-existing shape for any prompt that holds a run open (a Capture prompt does the same); not grown into this round.
- An early draft added a `try/finally` around `api.executeChoice`'s variable clear; review found a counter-scenario (a macro catching a cancelled sub-choice would lose its own shared variables), so released semantics are preserved exactly.
- Suite: 4815 tests green, svelte-check/eslint clean, production build + Astro docs build pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Multi-choice execution now correctly waits for picker completion and cleanly reports cancellation/errors in non-interactive and remote runs.
  * Completion handling for choice picking is now settled reliably when the picker ends, including empty-folder/backdrill cases.
  * Choice creation/configuration no longer risks losing a newly created choice during mid-configuration updates.
  * Provider model syncing matching is more consistent to ensure synced model merges apply correctly.

* **Documentation**
  * Clarified that AI prompt templates and bundled markdown notes can execute inline JavaScript and expand macros/templates on every AI run.
  * Reworked AI Assistant setting semantics (live vs copied defaults) and clarified inline script usage in AI Assistant prompt templates and Capture formats.
  * Expanded package/choices guidance to reflect the “runnable as code” capability behavior.


chhoumann:
@coderabbitai The duplicate completion-settle shape is extracted to a single `settleCompletion` helper in 835560c2 - both terminal branches route through it now, so the two call sites can't drift.

PR #1681, opened 2026-08-23
fix(date-picker): stabilize month navigation

### LINKED ISSUE #1680: [BUG] Date picker: hardcoded English UI strings, and the month header shifts with long month names
# Date picker: hardcoded English UI strings, and the month header shifts with long month names

## Summary

The date picker rendered for `date` inputs has three UI strings hardcoded in English, and its
header layout breaks when localized month names are wider than the English ones. Both problems only
surface for users on a non-English locale, which is probably why they have gone unnoticed.

Two separate problems in the same ~20 lines of code — happy to split into two issues if you prefer.

---

## 1. Hardcoded English strings

The calendar's month and weekday names go through `Intl.DateTimeFormat(undefined, …)` and localize
correctly — on a Polish system they render as `październik 2026`, `pon.`, `wt.` and so on. The chrome
around them does not:

| element | string | class |
| --- | --- | --- |
| action button | `Today` | `.qa-date-picker__action` |
| action button | `Clear` | `.qa-date-picker__action` |
| time field label | `Time` | `.qa-date-picker__time-label` |
| nav button `aria-label` | `Previous month` | `.qa-date-picker__nav` |
| nav button `aria-label` | `Next month` | `.qa-date-picker__nav` |

From the built `main.js` (2.22.0):

```js
const todayButton = actions.createEl("button", { cls: "qa-date-picker__action", text: "Today" });
todayButton.type = "button";
const clearButton = actions.createEl("button", { cls: "qa-date-picker__action", text: "Clear" });
clearButton.type = "button";
```

The result is a calendar that reads half Polish, half English.

This is easy to misdiagnose from the outside, so it may be worth a note in the docs: the widget looks
enough like the native Chromium `<input type="date">` popover that people assume it is one and go
hunting for a browser locale setting. I spent a while verifying that Obsidian's `locales/pl.pak`
contains the full set of translated calendar strings and that the renderer runs with a Polish
`ResourceBundle` before realising the picker is QuickAdd's own DOM.

### Scope

The same modal carries a few more English-only strings — `" (optional)"` appended to labels,
`"Skip (leave empty)"` in dropdowns, `"Provide inputs"` as the modal title, `Aliases (N)` on the
alias disclosure. I could not find a string table or any locale lookup for UI text in the bundle, so
this may be worth scoping as general i18n groundwork rather than a one-off fix for three labels.

### Workaround (for anyone who finds this issue first)

A CSS snippet, scoped to the picker:

```css
.qa-date-picker__actions .qa-date-picker__action { font-size: 0; }
.qa-date-picker__actions .qa-date-picker__action::after { font-size: 0.75rem; }
.qa-date-picker__actions .qa-date-picker__action:nth-child(1)::after { content: "Dzisiaj"; }
.qa-date-picker__actions .qa-date-picker__action:nth-child(2)::after { content: "Wyczyść"; }

.qa-date-picker__time .qa-date-picker__time-label { font-size: 0; }
.qa-date-picker__time .qa-date-picker__time-label::after { content: "Godzina"; font-size: 0.75rem; }
```

It works visually, but it leaves the English text in the accessibility tree and depends on button
order, so it is a stopgap rather than a fix.

---

## 2. The month header shifts on every navigation

`.qa-date-picker__header` is a flex container with `justify-content: space-between`, and
`.qa-date-picker__label` has no `min-width: 0`. A flex item does not shrink below its min-content
width by default, so a long month name pushes the header wider than the column it sits in and drags
the `›` button along with it.

Current CSS:

```css
.qa-date-picker__header {
  display: flex;
  align-items: center;
  justify-content: space-between;
  gap: .5rem;
  margin-bottom: .25rem;
}
.qa-date-picker__label { font-weight: 600; }
```

In English the effect is mild — `May 2026` vs `September 2026`. In Polish the width roughly doubles
(`maj 2026` vs `październik 2026`), so the navigation arrows visibly jump sideways on every click,
which makes clicking through several months awkward.

One declaration fixes it:

```css
.qa-date-picker__label {
  flex: 1;
  min-width: 0;        /* the actual fix */
  text-align: center;
}
```

`flex: 1` and `text-align: center` are cosmetic — they keep the label centred between the arrows.
`min-width: 0` is the part that stops the overflow.

---

## Steps to reproduce

1. Set the OS locale to a language whose month names vary a lot in length (e.g. `pl-PL`).
2. Trigger any QuickAdd choice that renders a `date` input.
3. Month and weekday names are localized; `Today`, `Clear` and `Time` are English.
4. Click the next-month arrow repeatedly — the arrow moves horizontally as the month name changes width.

## Environment

- QuickAdd 2.22.0
- Obsidian 1.13.7, desktop (Electron), installer and app both current
- Windows 11, OS display language `pl-PL`, Obsidian UI language English

Both issues are visible directly in the shipped bundle rather than being environment-dependent: the
three strings are literals in `main.js`, and the missing `min-width: 0` is in `styles.css`. Note that
my own vault now has the CSS workaround above enabled, so reproducing on my machine requires turning
that snippet off first.

## Related

- #288 — the closed feature request that introduced the picker
- I could not find an existing open issue about localization of the picker; apologies if I missed one.

chhoumann:
Keep the previous and next month controls stationary when the localized month label needs more horizontal space.

The label now uses the available space between the controls, can shrink in constrained layouts, and keeps its text centered. The controls were already stable at the date picker's normal width in our reproduction, but they could move once the picker became unusually narrow.

This is intentionally limited to the layout problem. QuickAdd does not currently have an internationalization system, so the English-only date picker controls remain unchanged and the broader language request stays open.

Verified with Polish month names in a live Obsidian date prompt at both the default width and deliberately constrained widths. The full test suite, type-check, lint, and production build pass.

Refs #1680

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -1397,6 +1397,9 @@
 }
 
 .qa-date-picker__label {
+	flex: 1;
+	min-width: 0;
+	text-align: center;
 	font-weight: 600;
 }
 


chhoumann:
Superseded because GitHub's issue-linked branch marked #1680 for closure, while this change addresses only the layout portion.

PR #1685, opened 2026-08-23
test(e2e): seed files through the vault API

### LINKED ISSUE #1669: e2e: residual flake class post obsidian-e2e 0.10.0 - vault sync stalls under host load, short waits vs ~5s recovery latency
Tracking issue for a residual e2e flake class observed after obsidian-e2e 0.10.0 (recoverable dispatch, chhoumann/obsidian-e2e#25 / #26) landed. Filing so future failures get attributed correctly instead of being lumped in with the (fixed) 30s transport hangs.

## Observations (2026-08-13, ~00:00-00:20)

Fresh-launch full-suite loops against registry obsidian-e2e@0.10.0: 3 failures in 10 runs. Same machine, minutes apart, a 2-run spot check on the old 0.9.0 baseline (pre-cleanup commit 146e9b02) was green - too small a sample to separate version from environment (P ≈ 0.5 of 2 green runs at the observed failure rate). Crucially, the machine was under heavy interactive load the whole window (load average ~13, foreground applications dominating CPU), while the previous day's 19 consecutive linked-build runs on a quiet machine were all green.

Failure signatures - none are the old `ObsidianCommandTimeoutError` hangs, and zero `ObsidianCommandDispatchError`s appeared:

1. **Vault-to-disk sync stalls (runs 2, 3):** files written to disk by the sandbox were not visible via `app.vault.getAbstractFileByPath` for >10s (seedFile waits exhausted: T05/T07 at 10.8s/10.6s vs a 10s budget); in one case the reverse direction - a plugin-side vault write's content not observable on disk within 10s (A04). Failure artifacts show a healthy idle app: zero runtime errors, zero console messages, zero notices, and the artifact capture itself ran ~10 CLI commands successfully seconds after the failure. The command channel was fine; file-event delivery/indexing was not.
2. **beforeAll hook timeout at 30s (run 8):** `file-exists-behavior` harness setup exceeded vitest's 30s hook budget; `qa` stayed undefined and the afterAll TypeError followed.

## Two candidate mechanisms (not mutually exclusive)

- **Environment:** a backgrounded Electron app on a loaded macOS host gets deprioritized; both fsevents delivery (vault indexing) and the CLI's main-to-renderer bridge degrade. Fits the healthy-but-stale artifacts and the quiet-machine 19/19.
- **Recovery-latency exposure (0.10.0-specific):** recoverable dispatch converts a lost CLI reply from a rare 30s hang into a ~5s recovery. Inside a 10s `waitFor`, TWO losses now exhaust the budget silently (`waitForValue` swallows per-iteration errors), producing exactly the observed ~10-11s failure shape without any visible transport error. Under load, loss rates spike, so short waits are the new weak point. This is a real trade (frequent-short vs rare-catastrophic) even if the environment triggered it.

## Actions

- [ ] Clean A/B on an idle machine: >=10 fresh-launch runs per arm (0.9.0 @ 146e9b02 vs 0.10.0 @ master). If 0.10.0 fails on a quiet machine, instrument `src/core/dispatch.ts` recovery events (attempt timeouts, poll recoveries) via a linked build and correlate with the failing waits.
- [ ] chhoumann/obsidian-e2e#28 item 1 (in-app grace period) cuts recovery latency to ~2.5s, which halves the two-losses-per-10s-wait exposure regardless of the A/B outcome - prioritize it.
- [ ] Suite hardening worth considering independently: seedFile budgets sized as (2x recovery latency + indexing time); or seed test files through the vault API (`app.vault.create` via evalJson) instead of raw fs writes when a test needs immediate indexing, removing the fsevents dependency entirely.

Artifacts from the failing runs are in `.obsidian-e2e-artifacts/` (t05/t07/a07/a08 dirs dated 2026-08-13 00:02).

chhoumann:
## Why

Issue #1669 tracks a residual e2e flake after obsidian-e2e 0.10.0. Seeded files written with `sandbox.write` plus `waitForContent: true` can sit on disk while `app.vault.getAbstractFileByPath` stays null.

`VaultApi.write` uses Node `writeFile`. `waitForContent` then re-reads those same bytes. It never asks the vault index. PR #1639 already hit this race and added a 10s `evalJson` poll. That poll still depends on fsevents, and each iteration rides recoverable `exec()`. A lost CLI reply costs `DISPATCH_ATTEMPT_TIMEOUT_MS` (5s) inside the 10s `WAIT_OPTS` budget. Two losses exhaust the wait with no transport error, which matches the T05/T07 artifacts.

QuickAdd then reads templates and collision targets through `getAbstractFileByPath`. A seed that is only on disk is invisible to those paths.

## Scope

`seedVaultFile` in `tests/e2e/e2eVault.ts` creates or overwrites the file with `app.vault.create` / `app.vault.modify` via `evalJsonAsync`. Parent folders use `createFolder`. The helper returns only after `getAbstractFileByPath` sees the file.

Local `seedFile` / `seedTemplate` wrappers in the e2e suites now call it. Post-run `waitForContent` / `waitForExists` assertions stay on disk. Those wait for plugin writes, not seeds.

This does not implement obsidian-e2e#28 (in-app grace period). It does not change `verify()`'s direct `obsidian --help` transport.

## Tradeoffs

A shared helper beats copying the eval into each suite. Putting vault create into obsidian-e2e's `VaultApi` would break that package's "vault stays filesystem-only" rule, so the consumer owns the index-visible seed.

## Blast Radius

Test-only. No plugin behavior, settings, or release artifacts change. The next person who adds an e2e file should seed through `seedVaultFile` when the test later calls `getAbstractFileByPath`.

## Verification

Isolated worktree vault `quickadd-quickadd` on Obsidian 1.13.7.

Before the change, a 5-loop full suite at load ~8 reproduced a different flake on loop 4. `apply-template-to-active-note` `beforeAll` died with `ObsidianCommandTimeoutError: Command timed out after 30000ms: obsidian --help`. That is `verify()` on the direct transport, not the vault-index class. Loops 1-3 were 49/49. A 20-iteration fs-then-index probe stayed under 182ms, so the 10s index stall did not fire in this window.

After the change, three consecutive full-suite runs were 49/49 (27.6s, 27.9s, 38.5s). `pnpm run lint` and `pnpm run test` were 4915 passed / 41 skipped. `dev:errors` was clean after the instance recovered.

- [x] PR title follows Conventional Commits
- [x] Linked related issue
- [x] No release or migration impact

Fixes #1669

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test setup for creating and indexing vault files.
  * Added shared handling for creating required folders, reconciling existing folders, and updating test files.
  * Updated template, note, file-opening, formatting, and scorecard scenarios to use consistent vault seeding.
  * Added coverage for folder creation, existing files, reconciliation, errors, and file updates.
  * Simplified template lookups while preserving existing not-found error behavior.

### DIFF
--- a/tests/e2e/apply-template-to-active-note.test.ts
+++ b/tests/e2e/apply-template-to-active-note.test.ts
@@ -39,12 +39,7 @@ async function applyTemplate(
 
 	return obsidian.dev.evalJsonAsync<ApplyResult>(`(async () => {
 		try {
-			// The vault index can lag behind sandbox writes; poll for the note.
-			let file = null;
-			for (let attempt = 0; attempt < 50 && !file; attempt++) {
-				file = app.vault.getAbstractFileByPath(${JSON.stringify(notePath)});
-				if (!file) await new Promise((resolve) => setTimeout(resolve, 100));
-			}
+			const file = app.vault.getAbstractFileByPath(${JSON.stringify(notePath)});
 			if (!file) throw new Error("note not found: " + ${JSON.stringify(notePath)});
 			const leaf = app.workspace.getLeaf(false);
 			await leaf.openFile(file);
--- a/tests/e2e/apply-template-to-active-note.test.ts
+++ b/tests/e2e/apply-template-to-active-note.test.ts
@@ -1,6 +1,6 @@
 import { beforeAll, describe, expect, it } from "vitest";
 import type { ObsidianClient, SandboxApi } from "obsidian-e2e";
-import { createQuickAddE2EHarness, PLUGIN_ID } from "./e2eVault";
+import { createQuickAddE2EHarness, PLUGIN_ID, seedVaultFile } from "./e2eVault";
 
 // ---------------------------------------------------------------------------
 // Constants & types
@@ -19,12 +19,9 @@ const getContext = createQuickAddE2EHarness("apply-template");
 // Helpers
 // ---------------------------------------------------------------------------
 
-/** Write a file into the sandbox and wait for Obsidian to index it. */
 async function seedFile(sandbox: SandboxApi, name: string, content: string) {
-	await sandbox.write(name, content, {
-		waitForContent: true,
-		waitOptions: WAIT_OPTS,
-	});
+	const { obsidian } = getContext();
+	await seedVaultFile(obsidian, sandbox, name, content);
 }
 
 /**
--- a/tests/e2e/e2eVault.ts
+++ b/tests/e2e/e2eVault.ts
@@ -6,6 +6,7 @@ import {
 } from "obsidian-e2e";
 import type { ObsidianClient, SandboxApi, VaultRunLock } from "obsidian-e2e";
 import { createPluginHarness } from "obsidian-e2e/vitest";
+import { seedVaultFileInApp } from "./seedVaultFileInApp";
 
 export const PLUGIN_ID = "quickadd";
 
@@ -53,31 +54,7 @@ export async function seedVaultFile(
 ): Promise<string> {
 	const vaultPath = sandbox.path(relativePath);
 	const createdPath = await obsidian.dev.evalJsonAsync<string | null>(
-		`(async () => {
-			const path = ${JSON.stringify(vaultPath)};
-			const content = ${JSON.stringify(content)};
-			const segments = path.split("/").filter(Boolean);
-			segments.pop();
-			let folder = "";
-			for (const segment of segments) {
-				folder = folder ? folder + "/" + segment : segment;
-				if (!app.vault.getAbstractFileByPath(folder)) {
-					try {
-						await app.vault.createFolder(folder);
-					} catch (error) {
-						if (!app.vault.getAbstractFileByPath(folder)) throw error;
-					}
-				}
-			}
-			const existing = app.vault.getAbstractFileByPath(path);
-			if (existing) {
-				await app.vault.modify(existing, content);
-			} else {
-				await app.vault.create(path, content);
-			}
-			const file = app.vault.getAbstractFileByPath(path);
-			return file ? file.path : null;
-		})()`,
+		`(${seedVaultFileInApp.toString()})(app, ${JSON.stringify(vaultPath)}, ${JSON.stringify(content)})`,
 	);
 	if (!createdPath) {
 		throw new Error(`Failed to seed vault file: ${vaultPath}`);
--- a/tests/e2e/e2eVault.ts
+++ b/tests/e2e/e2eVault.ts
@@ -4,7 +4,7 @@ import {
 	resolveObsidianEnvOptions,
 	verifyVaultPath,
 } from "obsidian-e2e";
-import type { ObsidianClient, VaultRunLock } from "obsidian-e2e";
+import type { ObsidianClient, SandboxApi, VaultRunLock } from "obsidian-e2e";
 import { createPluginHarness } from "obsidian-e2e/vitest";
 
 export const PLUGIN_ID = "quickadd";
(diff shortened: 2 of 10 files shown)


chhoumann:
## Independent shipping verification

**Verdict: PASS+NOTES**

Independent of the PR author. Product code was not changed. Throwaway local probes were deleted. Instance stopped. Not merged.

### HEAD

`01bfaf6d0d08194c53ea87f55c72ed90ffb8578e` — matches the expected SHA. No drift vs `origin` head of `test/1669-e2e-seed-via-vault`.

Parent `master` (`67d014c3`) → head is three test-only commits:
- `e57721dd` test(e2e): seed files through the vault API
- `76958a0b` test(e2e): drop the apply-template index poll
- `01bfaf6d` test(e2e): index on-disk sandbox folders when seeding

`tests/e2e/e2eVault.ts` `seedVaultFile` sends `seedVaultFileInApp` through `evalJsonAsync` (`app.vault.create` / `modify`, parent folders via `createFolder` / `reconcileFolderCreation`). `apply-template-to-active-note.test.ts` no longer has the 50-attempt index poll.

### Live surface

- Isolated instance: Obsidian **1.13.7**, vault `quickadd-workspace`, plugin id `quickadd`
- Ready probe `quickadd:list` → `{"ok":true,"command":"quickadd:list","count":0,"choices":[]}`
- `eval code='Boolean(app.plugins.plugins.quickadd)'` → `true`
- HOME remapped to `OBSIDIAN_E2E_OBSIDIAN_HOME=/tmp/quickadd-obsidian-e2e/quickadd-workspace-d276d603164d/home`
- Vault selector used as a prefix

### Commands run

```bash
git rev-parse HEAD   # 01bfaf6d0d08194c53ea87f55c72ed90ffb8578e
git diff master...HEAD
.agents/setup        # Obsidian 1.13.7, xvfb, pnpm install, Linux e2e patch, build, provision
pnpm run --silent start:e2e-obsidian -- --print-env
export HOME="$OBSIDIAN_E2E_OBSIDIAN_HOME"
pnpm run --silent obsidian:e2e -- quickadd:list
# throwaway live probe via seedVaultFile + evalJsonAsync (deleted after)
pnpm exec vitest run --config vitest.e2e.config.mts \
  tests/e2e/file-exists-behavior.test.ts \
  tests/e2e/apply-template-to-active-note.test.ts
pnpm run --silent obsidian:e2e -- dev:errors
HOME=/home/ubuntu pnpm run --silent stop:e2e-obsidian
```

### 1. Immediate index after `seedVaultFile` (no wait)

`seedVaultFile` created `__obsidian_e2e__/seed-vault-live-probe-151471b1/immediate-1787475896498.md`. The next `evalJsonAsync` called `getAbstractFileByPath` with no poll/sleep:

```json
{
  "createdPath": "__obsidian_e2e__/seed-vault-live-probe-151471b1/immediate-1787475896498.md",
  "immediate": {
    "path": "__obsidian_e2e__/seed-vault-live-probe-151471b1/immediate-1787475896498.md",
    "indexed": true,
    "text": "LIVE_SEED_1787475896498"
  }
}
```

Present immediately. Content matched.

### 2. On-disk folder case

Natural `mkdir` on disk (same mechanism as `createSandboxApi`) is **already indexed** before the next eval on this Linux/inotify host:

```json
{
  "folderIndexedBefore": true,
  "diskExistsBefore": true,
  "fileIndexedAfter": true,
  "adapterHasReconcileFolder": true
}
```

`adapter.mkdir` in the same eval also returned the folder already indexed. A 15-iteration Node-`mkdir` + immediate `getAbstractFileByPath` loop was **0/15 unindexed**.

A fileMap-delete synthesis is **not** the sandbox-mkdir race: after the watcher has already seen the folder, `reconcileFolderCreation(folder, folder)` returns without error and leaves `getAbstractFileByPath` null. `createFolder` then throws `Folder already exists.` That is an artificial desync, not the pre-watcher window the helper targets. I could not safely produce the genuine race here.

### 3. Flake suites from #1669 (verbatim)

```
 RUN  v4.1.10 /workspace

 Test Files  2 passed (2)
      Tests  37 passed (37)
   Start at  09:05:55
   Duration  4.44s (transform 44ms, setup 0ms, import 102ms, tests 4.17s, environment 0ms)
```

Exit 0. `file-exists-behavior.test.ts` and `apply-template-to-active-note.test.ts` are green against the live instance.

(Earlier, a mis-filtered probe run also executed the rest of `tests/e2e/**/*.test.ts`: 9 files / 52 tests passed in 7.41s. Extra, not the required pair.)

### 4. `dev:errors` after the suites

Three `app.js` TypeErrors at **09:05:41** (14s **before** the official suites at 09:05:55):

```
09:05:41 Uncaught TypeError: Cannot read properties of undefined (reading 'lastIndexOf')
    at Pl (app://obsidian.md/app.js:1:546617)
    at t.onRaw (app://obsidian.md/app.js:1:2918312)
    ...
    at t.onChange (app://obsidian.md/app.js:1:1386601)

09:05:41 Uncaught TypeError: Cannot read properties of undefined (reading 'startsWith')
    at t.onRaw (app://obsidian.md/app.js:1:2898524)
    ...

09:05:41 Uncaught TypeError: Cannot read properties of undefined (reading 'lastIndexOf')
    at Al (app://obsidian.md/app.js:1:546549)
    at t.onRaw (app://obsidian.md/app.js:1:1153242)
```

No QuickAdd frames. Same `startsWith` message as the verifier diagnostic that called `adapter.reconcileFile` on a folder after deleting it from `fileMap`. Not a new plugin error from this PR. The required suites still passed afterward.

### 5. Stop

```
Stopped instance /tmp/quickadd-obsidian-e2e/quickadd-workspace-d276d603164d: terminated 8, killed 0, removed.
```

### Notes (non-blocking)

1. On this Linux/inotify surface the vault watcher indexes an on-disk folder before the next CLI eval. The `reconcileFolderCreation` follow-up in `01bfaf6d` is still the right consumer-side guard; it is just not observable here without corrupting `fileMap`.
2. `reconcileFolderCreation` does not rebuild an entry that the watcher already processed and that was then deleted from `fileMap`. Do not treat that as a substitute for the real race.
3. `dev:errors` above are from that diagnostic, not from QuickAdd or from `seedVaultFile`.

**PASS+NOTES:** live `seedVaultFile` is immediately visible to `getAbstractFileByPath`; both #1669 flake files are green; no new plugin errors.

PR #1732, opened 2026-09-02
test(date): assert calendar origin offsets in local time

chhoumann:
`addCalendarOffset` already snaps to local midnight via moment `startOf("day")`. The unit test asserted `toISOString().startsWith("2026-08-19")`, which is only true in UTC and UTC-west zones. East of UTC, local midnight on 19 Aug is still 18 Aug in UTC, so the test failed on a UTC+2 machine while CI (UTC) stayed green.

No production change. The DATE origin path formats with moment in local time (`getDate`), so users in Copenhagen, New York, and Sydney already got the intended calendar day, including across the 2026 DST switches those zones actually observe.

Assert local Y-M-D H:m instead, and cover week offsets that cross EU, US, and Australia/Sydney DST boundaries.

Verified `pnpm run test` (5250 passed) under `TZ=UTC`, `TZ=Europe/Copenhagen`, `TZ=America/New_York`, and `TZ=Australia/Sydney`.

## Summary by CodeRabbit

* **Tests**
  * Expanded calendar offset coverage across EU, US, and Australia/Sydney time zones.
  * Added validation for daylight-saving transitions and local-midnight behavior.
  * Updated assertions to verify local calendar components for week and year offsets.

### DIFF
--- a/src/utils/resolveDateOrigin.test.ts
+++ b/src/utils/resolveDateOrigin.test.ts
@@ -21,13 +21,42 @@ beforeEach(() => {
 	vi.setSystemTime(new Date("2026-08-26T15:30:00"));
 });
 
+// Local Y-M-D H:m. startOf("day") is local midnight, so toISOString() is the
+// previous UTC day east of UTC.
+const localDay = (d: Date) => [
+	d.getFullYear(),
+	d.getMonth() + 1,
+	d.getDate(),
+	d.getHours(),
+	d.getMinutes(),
+];
+
 describe("addCalendarOffset", () => {
-	it("moves by week and month from the frozen now", () => {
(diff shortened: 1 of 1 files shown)


chhoumann:
CodeRabbit's nit is right that CI never sets `TZ=Europe/Copenhagen` (or New York / Sydney). I'm leaving that out of this PR.

The assertion is local Y-M-D H:m, which is the same calendar day in every zone, including across those DST dates. Reintroducing `toISOString()` would still pass in UTC and in `America/Los_Angeles` (the platform-test job), so a TZ matrix would be the way to lock the original UTC+ failure in CI. That is a follow-up, not a blocker: production `addCalendarOffset` is unchanged, Codex had no findings, and `pnpm run test` was green under all four zones before this went up.

